PR #10, opened 2020-01-09
fix(fclasses) Design transformer does not update passthroughProps when its props are changed

beliayeu:
## Changes
* fixed an issue with updating design transformer props

## Test Instructions
* open a product page on test site
* click bv placeholder
* enter "BV Product External ID" (update product external id if it is already added)
* click save

actual: bv placeholder content is updated
expected: bv placeholder content is not updated


beliayeu:
per @wodenx, @e2thex 

we decided to have the transformer function return different props for constructor and render - eg
```
const [ fixedProps, dynamicProps ] = transformer(props);
```

then we'd call it twice - once in constructor to get passthroughProps,

```
const { transformer,...rest } = props;
const [ fixedProps ]  = transformer(props, ['component']));
this.fixedProps = fixedProps;
```

and once in render to get remaining props and render would be:

```
const { transformer, component: Component , ...rest } = this.props;
const [ , dynamicProps ] = transformer(...rest);
<Component {...this.fixedProps} {...dynamicProps}>
```

beliayeu:
@wodenx, @e2thex, per my understanding one of the concerns was that we want to avoid invoking the transformer during render phase. is it no longer a concern?

packages/fclasses/src/Design.tsx
```diff
@@ -141,20 +141,22 @@ type AlterPropsProps<P> = {
...
   AComponent: ComponentType = React.Fragment;
 
-  passthoughProps = {};
+  transformer: Function;
 
   constructor(props:AlterPropsProps<P>) {
     super(props);
-    const { transformer, acomponent, ...passthoughProps } = props;
-    this.passthoughProps = transformer(passthoughProps);
+    const { transformer, acomponent } = props;
```
wodenx: Discussed this with @e2thex  we decided to have the transformer function return different props for constructor and render - eg
```
const [ fixedProps, dynamicProps ] = transformer(props);
```
then we'd call it twice - once in constructor to get passthroughProps,
```
const { transformer,...rest } = props;
const [ fixedProps ]  = transformer(props, ['component']));
this.fixedProps = fixedProps;
```

and once in render to get remaining props and render would be:
```
const { transformer, component: Component , ...rest } = this.props;
const [ , dynamicProps ] = transformer(...rest);
<Component {...this.fixedProps} {...dynamicProps}>
```

beliayeu: @wodenx, @e2thex, per my understanding one of the concerns was that we want to avoid invoking the transformer during render phase. is it no longer a concern?

e2thex: @beliayeu I think it is fine because we are not using the result of the HOC in the render

packages/fclasses/src/Design.tsx
```diff
@@ -136,39 +136,52 @@ export const remove = <P extends React.HTMLAttributes<HTMLBaseElement>> () => (p
...
+    this.Component = Component;
+    this.transformer = transformer;
+    const { fixedProps } = transformer(rest);
+    this.fixedProps = fixedProps;
   }
 
   render() {
-    return <this.AComponent {...this.passthoughProps as P} />;
+    const { transformer, Component, ...rest } = this.props;
+    const { passthroughProps } = this.transformer(rest);
```
wodenx: So after discussing this with @beliayeu , I think it's better to pass 2 functions rather than a single function that returns both sets of props -- otherwise we're calling applyDesign needlessly on every render (even though we through away the result).  Can make the interface clean by having, eg, 'transformer.onConstructor(props)` and `transformer.onRender(props)` or the like.

e2thex: I was thinking the same thing will do

e2thex: @wodenx ok I made the change I also do the following
* Change the test to be testing designable
* Added a test on designable to test that change the design prop do not change the output.
* Clean up the typing a bit so that is saner and more precise

e2thex:
fixed #10

e2thex:
fix #17

PR #18, opened 2020-01-09
feat(richtext): Refactor the RichText API

e2thex:
## Changes
* Makes the RichText Component take a design prop following the DesignAPI
* Change internal to have use components and not richTextItems (removed the RichTextItems)
* Create functionality in RichText that makes it so that we have a default for component with default configuration
* Update documantation 

## Test Instructions

* In general regression testing of the richtext editor needs to be done. (nothing in simple and basic should have changed)
* The technical review should play around with using the following function in the test site in the design that is pass to the richtext editor
** asBlock (should expect that inline items can be set inside of it)
** withButton (should show a button that toggles the formating)
** withShortcutKey (should allow the ctrl-KEY should toggle the format)
** other style related HOC should style appropriately
** startWith (should change the underline component)
There are examples to review of all of this in the full-featured editor

## Related Issues

Closes #10 

BREAKING CHANGE: this changes the API for the richtext Editor


examples/test-site/src/data/pages/gallery-final/withSimpleEditor.tsx
```diff
@@ -24,13 +24,17 @@ import {
...
 import asEditor from '../../../components/Editors/asEditor';
-import { asLink } from '../../../components/Elements.token';
+
+const asBold = withComponent(Strong);
+const asItalic = addClasses('');
+const asUnderline = addClasses('underline');
+const asLink = flow(asBodilessLink(), addClasses('text-blue-700 underline'));
 
 const simpleDesign = {
-  Bold: withComponent(Strong),
```
e2thex: @beliayeu why are we not pulling this from token and also can we just not use this file and use the site simple editor?

beliayeu: @e2thex, withSimpleEditor.tsx is created for gallery page which is the page that you build in the tutorial (https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-documentation/doc/About/SiteBuildBasics.md#rich-text)
it is intended to demonstrate how you would create a custom editor

Issue #39, opened 2020-01-15
Documentation of Bodiless-ui Changes


malachye:
- Move Technical Details to README.md of bodiless-ui

PR #40, opened 2020-01-15
docs(bodiless-ui): add technical details

malachye:
- Add technical details to bodiless-ui readme.md

**Related Issues:**
#39


packages/bodiless-ui/README.md
```diff
@@ -110,3 +110,45 @@ export default class Example extends Component {
   }
 }
 ``` 
+## Technical Details
```
hvanyo: Can we also add a section right before this.

## Bodiless classes.
Bodiless uses tailwind for its styling of bodiless editor and a site built with this tool also uses tailwind for the site styling, so that leads to two tailwind config in use.  

Bodiless tailwind config resides in packages/bodiless-ui/bodiless.tailwind.config.js and we have prefaced it with the prefix bl- to isolate styling needed for the Bodiless editor.     The sites tailwind.config.js resides at site level (i.e. examples/test-site or examples/starter) and this is also built.   With that we do introduce in two css file within the edit mode, a prefixed (bs-) for bodiless editor and a non-prefixed for the site.   The static mode would only serve the non-prefixed (at this point in time this is future enhancement to remove the bodiless editor from static build.)

So recommended guideline is to prefix any bodiless editor styling with bl- and any site level without the prefix and continue this separation.

malachye: Added section.

PR #52, opened 2020-01-17
fix(gatsby-theme-bodiless): Typed chars disappear sporadically

### LINKED ISSUE #14: Last typed chars disappear sporadically; some entered data is lost after refreshing a page
## Description

Last typed chars disappear sporadically; some entered data is lost after refreshing a page

### Steps to reproduce

1. open /list page
2. start adding the list items and fill them with text

### Expected result

no content loss happens

### Actual result

last chars in some items are lost after several seconds

## Root cause

* stale data come from Gatsby backend and overwrite client data

##  Background
1. once user makes content changes, browser locks items, to which the user made changes, and sends data to backend. the item is unlocked on timeout. current timeout value is 5 seconds. if user touches an item then lock timeout is reset.
2. backend saves the data from 1. to json files
3. gatsby watches json files and triggers site rebuild once a json file is changes
4. after site is rebuilt, the data is sent back to the client.

issue: the data on  step "4" can be stale. when the data from "4" is delivered to the browser and if the browser already released locks for the items, then the stale data overwrites more recent data. it looks for the user as some characters are disappeared.

## Options how to solve the stale issue
Option 1. Increase item lock timeout.
Option 2.  Do not expire item locks.
Option 3.  Do not refresh with changes made in same browser
Option 4. Verify data before accepting it.
Option 4a. verify data in browser by sending a network request to backend
Option 4b. verify data in gatsby-node.js

## Ideas how to improve simultaneous editing
* we can have a lock for each item on the server side. the client has to acquire the lock in order to make changes for an item. if an item is already locked on the server by a client, then the data originated from other clients is rejected.
* implement [Operational transformation pattern](https://en.wikipedia.org/wiki/Operational_transformation)

beliayeu:
## Changes
* fix for the problem reported in #14

### Bodiless backend changes
* adjusted bodiless data model. extended json data with metadata. the metadata has author field that contains uuid of a client that made last modification of the data. The client uuid is taken from http request header.

### Bodiless Gatsby theme changes
*  added client uuid to requests sent to backend. The client uuid is generated for MobX store and will change on page reload.
* MobX store rejects data if the data was originated from the same client.

## Test Instructions

see #14

## Related Issues

Fixes #14


packages/bodiless-backend/src/processPayload.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright © 2019 Johnson & Johnson
```
VancheeZze: Copyright © 2020

packages/bodiless-backend/src/processPayload.js
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const processPayload = req => {
+  const { body } = req;
+  let meta;
+  if (req.headers['x-clientid']) {
```
wodenx: I think we shoujd namespace this header - eg `x-bl-clientid`

packages/bodiless-backend/package.json
```diff
@@ -12,7 +12,8 @@
     "dotenv": "^8.2.0",
     "express": "^4.16.4",
     "formidable": "^1.2.1",
-    "tmp": "^0.0.33"
+    "tmp": "^0.0.33",
+    "uuid": "^3.3.2"
```
wodenx: Why do we need `uuid` here?

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStore.ts
```diff
@@ -72,15 +75,37 @@ class Item {
     }
   }
 
+  private shouldAccept(metaData: any) {
```
wodenx: We should type the metadata.

wodenx:
@beliayeu This looks good - if it does, in fact, solve the problem we should consider a next step which keeps the metadata out of git.  As discussed, the best option for this is to have the backend write data+metadata to a separate file which is git-ignored (eg `{node$path}.changed.json`), then either:
- exclude these files from gatsby query results, but check for their existence for all nodes in onCreateNode, and if they exist, use their data in place of the original file.  Have to use *all* their data (not just metadata) 
- include these files in gatsby query results, but post-process results to replace original with changed, either on server side (in gatsby-node, maybe use `sourceNodes` hook?) or in client (in GatsbyMobxStore::updateData)

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStore.ts
beliayeu: fixed

packages/bodiless-backend/package.json
beliayeu: removed, it was not needed

packages/bodiless-backend/src/processPayload.js
beliayeu: makes sense, renamed the header

packages/bodiless-backend/src/processPayload.js
beliayeu: fixed

beliayeu:
@wodenx, @VancheeZze, changes since your last review
* show alert to the user when the user leaves the pages and there are pending requests
* increased mobx debounced time to 2 seconds in order to mitigate the issue when a network request is delivered to the backend before its predecessor
* added extended logging to bodiless backend leveraging morgan
cc: @TatsianaMiraniuk

packages/bodiless-backend/README.md
```diff
@@ -10,5 +10,7 @@ You can start backend server by running `node ./path/to/server.js`. Backend serv
 * `BODILESS_BACKEND_DATA_FILE_PATH` - Where to write the content json files. Defaults to `./src/data`.
 * `BODILESS_BACKEND_STATIC_PATH` - Where to write uploaded static assets. Defaults to `./static`.
 * `BODILESS_BACKEND_COMMIT_ENABLED` - Whether or not commits are enabled. Defaults to `0` for `production` and `development` environments and `1` for `test/*` and `changeset/*` branches.
+* `BODILESS_BACKEND_COMMIT_ENABLED` - Whether or not commits are enabled. Defaults to `0` for `production` and `development` environments and `1` for `test/*` and `changeset/*` branches.
```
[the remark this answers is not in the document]
beliayeu: fixed

packages/bodiless-backend/README.md
```diff
@@ -10,5 +10,6 @@ You can start backend server by running `node ./path/to/server.js`. Backend serv
 * `BODILESS_BACKEND_DATA_FILE_PATH` - Where to write the content json files. Defaults to `./src/data`.
 * `BODILESS_BACKEND_STATIC_PATH` - Where to write uploaded static assets. Defaults to `./static`.
 * `BODILESS_BACKEND_COMMIT_ENABLED` - Whether or not commits are enabled. Defaults to `0` for `production` and `development` environments and `1` for `test/*` and `changeset/*` branches.
+* `BODILESS_BACKEND_EXTENDED_LOGGING_ENABLED` - Whether or not extended logging is enabled. Defaults to `0`.
```
VancheeZze: I would also add BODILESS_BACKEND_EXTENDED_LOGGING_ENABLED to `.env.site` to both test-site and starter in the `/examples/` folder.

```
# The Extended Logging is disabled by default.
# To enable it - uncomment the following line.
# BODILESS_BACKEND_EXTENDED_LOGGING_ENABLED ='1'
```

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStore.ts
```diff
@@ -143,9 +170,13 @@ export default class GatsbyMobxStore {
...
   data: any;
 
+  storeId: string;
+
   constructor(nodeProvider: DataSource) {
     this.setNodeProvider(nodeProvider);
-    this.client = new BackendClient();
+    this.storeId = v1();
+    this.client = new BackendClient({ clientId: this.storeId });
+    addPageLeaver(this.client.getPendingRequests.bind(this.client));
```
wodenx: @beliayeu I think there's an issue here, sorry I didn't catch it before.  We now have a 2s debounce on updates before a post is issued -so if you refresh the page within that 2 seconds, there won't have been a post yet and so there won't be any pending requests, and the changes you made will be lost...

beliayeu: @wodenx, refined the logic in https://github.com/johnsonandjohnson/Bodiless-JS/pull/52/commits/e003a9aa4b090661305f559ea16d5c681546e3fe

packages/bodiless-backend/README.md
beliayeu: @VancheeZze, fixed in https://github.com/johnsonandjohnson/Bodiless-JS/pull/52/commits/3b081c862f77a0e83c0773c9bf91fbb8d663eb82

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStore.ts
```diff
@@ -103,6 +93,13 @@ class Item {
     }
   }
 
+  private setStatus(newStatus: ItemStatus) {
+    if (newStatus === ItemStatus.Clean && this.status !== ItemStatus.Flushing) {
```
wodenx: Add a doc here describing the purpose of the check, eg:
```suggestion
    if (newStatus === ItemStatus.Clean && this.status !== ItemStatus.Flushing) {
      // If an update happens while flushing, the status will be set to dirty. In this case
      // we don't want to reset it to clean.
```

or - maybe we'd be better served here by a simple state machine, eg:
```
enum ItemStateEvent {
  UpdateFromBrowser,
  BeginPostData,
  EndPostData,
}

private updateState(event: ItemStateEvent) {
  switch (event) {
    case ItemStateEvent.UpdateFromBrowser:
      this.state = ItemState.Dirty;
      break;
    case ItemStateEvent.BeginPostData:
      this.state = ItemState.Flushing;
      break;
    case ItemStateEvent.EndPostData:
      this.state = this.state === ItemState.Dirty ? ItemState.Dirty : ItemState.Clean;
      break;
  }
}
```
To me this is clearer, but up to you if you want to take the time to refactor it.

beliayeu: @wodenx, thanks for your suggestion.
addressed in https://github.com/johnsonandjohnson/Bodiless-JS/pull/52/commits/b91098c4c2e7eb2da36a52190875895e13c185a9

packages/gatsby-theme-bodiless/src/dist/addPageLeaver.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+const addPageLeaver = (getPendingRequests: () => any[]) => {
+  if (typeof window !== 'undefined') {
+    window.addEventListener('beforeunload', e => {
+      if (getPendingRequests().length > 0) {
+        // Cancel the event
+        e.preventDefault();
+        // Chrome requires returnValue to be set
+        e.returnValue = 'Are you sure you want to leave?';
```
VancheeZze: This may not work in chrome and some other browsers including mobile browsers, but I don't think there is a better way to prevent from leaving the page.

PR #57, opened 2020-01-17
feat(fclasses): Change startWith so that it does not replace the whol…

e2thex:
…e item but instead just the starting Component

## Changes

Change the functionality of startWith, to not replace the component where it is but to change the component that apply design is going to start with.

## Test Instructions

Use startWith after applying HOC and see that the component is changed but the HOC is still applied.

## Related Issues

BREAKING CHANGE: startWith functionality is changed


packages/fclasses/src/Design.tsx
```diff
@@ -80,6 +80,25 @@ const withDisplayName = <P extends Object> (name: string) => (Component: Compone
   const newMeta = mergeWith({}, Component, { displayName: name });
   return Object.assign(WithDisplayName, newMeta);
 };
+const designContextDefault = undefined as undefined | ComponentType<any>;
+const DesignContext = React.createContext(designContextDefault);
+export const replaceable = <P extends object> (Component:ComponentType<P>) => (props:P) => {
```
wodenx: Will this have a proper displayName?

examples/test-site/src/components/Editors/index.tsx
```diff
@@ -48,7 +50,7 @@ const simpleDesign = {
   SuperScript: asSuperScript,
 };
 const basicDesign = {
-  Bold: asBold,
+  Bold: flow(asBold, startWith(Em)),
```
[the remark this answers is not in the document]
e2thex: I need to fix this was just a test

- [ ] Remove EM

packages/fclasses/src/Design.tsx
e2thex: - [ ] Refactor to have proper displayName

PR #73, opened 2020-01-22
Fix(Richtext): URL link value and Header style resets after applying (#72)

### LINKED ISSUE #72: Full RTE: URL link value and Header style resets after applying a text alignment
### Steps to reproduce
1. Open Home page, http://0.0.0.0:8005/
2. Add some text in RichTextEditor
3. Make some word as a link and add URL value
4. Apply Text Alignment to the text (select any option for that from the left menu)
5. Check URL of the link

### Expected result
URL value retains after applying any style options.

### Actual result
URL value is lost (reset to default '#')

### Affected Version
0.0.37

dewen:
## Changes

Updated slate content node child getter logic so it's not saved in separate node file.

## Test Instructions

* Test edit feature on RichText component.
* Regression test on RichText content pages.

## Related Issues

Fixes #72


wodenx:
@dewen PR title has incorrect format - should be "fix(richtext): ..."

packages/bodiless-richtext/src/RichTextItemGetters.tsx
```diff
@@ -52,14 +56,15 @@ const SlateComponentProvider = (update:Function) => (
...
-      const nodeKey = data.nodeKey || newData.nodeKey;
-      // We might need to understand if we need a nodecollection
-      const contentNode = useNode<D>().node.child(nodeKey);
+      const getters = {
+        getNode: () => node.data.toJS(),
+        getKeys: () => ['slatenode'],
+      };
+      const actions = {
+        // tslint: disable-next-line:no-unused-vars
+        setNode: (path: string[], componentData: any) => update({ node, editor, componentData }),
```
wodenx: @dewen as we discussed, this will not work if the component itself has child nodes.  As it stands, any call to the "setNode" setter will overwrite the top level data object.

wodenx: key-value store, keyed by node path (eg "slatenode$outer$inner") would probably be simplest (this is what we do in the actual gatsbymobxstore.  so
```
setNode: (path, data) => update({ node, editor, componentData: { ...node.data, [path.join('$')]: data });
```
etc

dewen: toJS() added as
```
        setNode: (path: string[], componentData: any) => update({
          node,
          editor,
          componentData: {
            ...node.data.toJS(),
            [path.join('$')]: { ...componentData },
          },
        }),

```

Issue #50, opened 2020-01-17
Environment variables management


VancheeZze:
### User Story

- AS A Site Builder
- I WANT to build packages that have the values set up
- SO THAT I do not have to add more work. 

### Acceptance Criteria

AC1:
There is a mechanism by which any bodiless package can implement logic to set environment variables for both edit (.env.development) and static (.env.production) applications

AC2.
The logic so implemented will be invoked during the build (static) or start (edit).

AC3.
The site may specify absolute values and/or logic to set environment variables, and if so, site-level logic will override logic implemented in any package.

AC4: Documentation is updated with where the variables are updated

AC5: The linked bugs are retested and no longer an issue. 

AC6. Current default logic will be implemented in gatsby_theme_bodiless package

PR #83, opened 2020-01-27
docs(examples): Add "How To" generate env vars and restructure .env.site comments.

VancheeZze:
## Changes
- Add documentation on how to generate env variables to the example sites.
- Restructure `.env.site` comments style to follow the same pattern.
- Added `build:env-vars` script to the root package.json which leverages lerna to run `build:env-vars` script in all packages.

## Related Issues
Relates to: https://github.com/johnsonandjohnson/Bodiless-JS/issues/50

### DIFF
--- a/examples/test-site/.env.site
+++ b/examples/test-site/.env.site
@@ -1,8 +1,10 @@
-# Set to string 'true' to enable Git commits. Default is 'false' if not specified.
+# Git commits are disabled by default.
+# To enable it - uncomment the following line.
 # BODILESS_BACKEND_COMMIT_ENABLED='1'
 
-# Set to string 'true' to enable save. Default is 'true' if not specified.
-# BODILESS_BACKEND_SAVE_ENABLED='1'
+# The Backend Data Save is enabled by default.
+# To disable it - uncomment the following line.
+# BODILESS_BACKEND_SAVE_ENABLED='0'
 
 # Git commit paths used by the backend, in our case @bodiless/backend/server.js.
 # APP_GIT_PATH='.'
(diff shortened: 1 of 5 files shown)


examples/test-site/.env.site
```diff
@@ -1,30 +1,31 @@
...
+# The Google Fonts plugin is enabled by default.
+# To disable it - uncomment the following line.
 # GOOGLE_FONTS_ENABLED='0'
 
-# GTM is enabled on Bodiless example site by default and is rendered in the static site.
-# For GTM activation/disablement, you should set the following variables in the .env.production file that controls static site.
-
-# GTM Disablement
-# GTM is enabled by default. To disable GTM, uncomment the following line.
+# GTM is enabled by default.
```
hvanyo: GTM is disabled by default

examples/test-site/.env.site
```diff
@@ -1,30 +1,31 @@
...
+# To disable it - uncomment the following line
```
hvanyo: to enable it
next line set to = "1"

examples/test-site/.env.site
VancheeZze: Updated.

examples/test-site/.env.site
VancheeZze: Done

examples/test-site/.env.site
VancheeZze: I think it just this section is never gets updated. If we look into "Files Changed" then you will see the changes: https://github.com/johnsonandjohnson/Bodiless-JS/pull/83/files#diff-9d823bf4f52f8e81b8d295475c23c541R19

PR #105, opened 2020-02-05
feat(layouts): Component Switcher

### LINKED ISSUE #69: Component Switcher
## User Story
As a content editor
I want to be able to replace one component with another
So I have control over the content on my page.

## Acceptance Criteria

AC1.
There exists a "component switcher" component
Which adds a button to the menu
Allowing you to select what component it should render
Using the existing component selector.

AC2.
If content data exists for the first component
The new component gets the same data.

AC3.
If the new component has a different data schema
It will act as though it had no initial data
And the data for the original component will be ignored.

## Implementation Note

Implement by creating a special case for flexbox.

wodenx:
## Changes

- Adds the ability to switch out a component in the flexbox while retaining the same data.
- Refactor of flexbox to use more standard bodiless hoc integrations
- Causes insert item action to insert new item after the current item 
- Causes delete item action to focus the adjacent item rather than the flexbox itself (unless flexbox is empty).

## Test Instructions

- On homepage of test site click on one of the touts and then click the "switch" button -- replace with another, similar, tout.  Verify that the content (for whatever fields exist on both touts) remains the same.
- Delete a tout.  Verify that the other tout has focus
- Delete the other tout, verify that the flexbox has focus (+ button avilable).
- Add back some touts
- insert a tout when first tout is focused - verify that new tout is added in the middle.

## Related Issues

Closes #69


hvanyo:
Issue -- flexbox page if empty doesn't get a + to add more compoments

STR:

1. Visit http://localhost:8005/flexbox
2. try click in the 'right' place and get + to add a component

OR

1. visit http://localhost:8005/
2. Remove all touts
3. Click somewhere to take focus off flexbox
4. Click in the 'right' place to add new components to flexbox area.

ER
+ will appear
AR
+ will not appear

hvanyo:
Observation (no action needed or maybe document somewhere)
If we keep our components named with similar patterns, 
Tout has "Title"  "Body"   and Accordion has "TItle"  "Body" -- these "not like" components switch back & forth just fine.

I assume something that had an Image like a Tout -- would switch back & forth as nicely.. I don't have that example to try out.  

Another thought if the "Body" could switch out with Basic/Simple/Full RTE that might be kind of cool.  But that is beyond scope of this ticket.  but if it was easy..  :)

wodenx:
@hvanyo The bug you reported is fixed in https://github.com/johnsonandjohnson/Bodiless-JS/pull/105/commits/ee8744a70e8efc192ec30d2a598ccd15d6a03afb,  For some reason I can't reply to your comment (maybe because you didn't actually "start a review" when you left the comment?

wodenx:
@hvanyo 
> Another thought if the "Body" could switch out with Basic/Simple/Full RTE that might be kind of cool. But that is beyond scope of this ticket. but if it was easy.. :)
Did you try this?  I think it should "just work" as long as all editors use subsets of the same components.

wodenx:
@e2thex It says you left a review and comments but i don't see them.

hvanyo:
@wodenx 
I can't reply to your comment above so will quote

> Another thought if the "Body" could switch out with Basic/Simple/Full RTE that might be kind of cool. But that is beyond scope of this ticket. but if it was easy.. :)
> Did you try this? I think it should "just work" as long as all editors use subsets of the same components.

I did try and it didn't work. :(    I wanted to see behaviors when I replaced with non-similar components, so started replacing everything with anything.

hvanyo:
> @hvanyo The bug you reported is fixed in [ee8744a](https://github.com/johnsonandjohnson/Bodiless-JS/commit/ee8744a70e8efc192ec30d2a598ccd15d6a03afb), For some reason I can't reply to your comment (maybe because you didn't actually "start a review" when you left the comment?

Verified fixed

Issue #114, opened 2020-02-06
Preview Mode Session


hvanyo:
### Background/Description

As a Content Editor, I want to navigate the site in Preview mode without having to switch back and forth, so that I can continuously preview the site without interruption. 
 *  Issue to fix: Edit mode turns on when navigating from a page where edit mode is off.  It should stay off.
 * Need to store in Preview mode session

### Acceptance Criteria

GIVEN a User is in Preview mode
WHEN they select a link to navigate the site
THEN the Edit UI stays in preview mode when jumping to the different pages until the User has selected Edit mode again.

PR #117, opened 2020-02-07
feat(core): Preview Mode Session

VancheeZze:
## Changes
- Edit mode is now persisted with a sessionStorage

## Test Instructions

- Run `npm run start`
- Switch the edit mode off.
- Navigate off the page or refresh it. The edit mode should stay off.

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/114


packages/bodiless-core/src/PageEditContext/index.tsx
```diff
@@ -61,7 +61,7 @@ export class PageEditStore implements PageEditStoreInterface {
 
   @observable contextMenuOptions: TMenuOption[] = [];
 
-  @observable isEdit = true;
+  @observable isEdit = !!JSON.parse(typeof window !== 'undefined' ? window.sessionStorage.getItem('isEdit') || 'true' : 'true');
```
beliayeu: @VancheeZze, my recommendation
1. wrap JSON.parse by try/catch since we will get uncaught syntax error if the session storage item contains invalid json
2. move the logic to read from storage to a separate function for better maintainability/readability

VancheeZze: Done

packages/bodiless-core/src/SessionStorage.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable no-console */
```
beliayeu: @VancheeZze , please add copyright

VancheeZze: Done

PR #123, opened 2020-02-11
fix(psh): Ensure cypress cache folder is writable.

wodenx:
## Changes
- Set cypress cache directory location on writable volume on p.sh

## Test Instructions
- Activate edit environment and/or run `platform ssh bash ./platform.sh fresh`.  Verify that installation completes and app starts normally.

### DIFF
--- a/packages/bodiless-psh/resources/edit/edit.platform.app.yaml
+++ b/packages/bodiless-psh/resources/edit/edit.platform.app.yaml
@@ -57,3 +57,5 @@ variables:
     env:
         APP_VOLUME: '/app/volume'
         CHOKIDAR_USEPOLLING: true
+        # Ensure cypress cache is writable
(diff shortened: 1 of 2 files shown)


packages/bodiless-psh/resources/edit/edit.platform.app.yaml
```diff
@@ -57,3 +57,5 @@ variables:
     env:
         APP_VOLUME: '/app/volume'
         CHOKIDAR_USEPOLLING: true
```
hvanyo: @wodenx  Does this mean that these resources that pushed to sites will use cypress as well?

hvanyo:
Closing as won't fix from @wodenx  "that can be closed I fixed it a different way in the other pr" -> https://github.com/johnsonandjohnson/Bodiless-JS/pull/135

PR #129, opened 2020-02-11
feat(documentation): Support better control over IA

wodenx:
## Changes
- Modify documentation app 
  - Use js instead of json to define TOC's, enabling reuse of existing tocs.
  - Allow separation of docs into namespaces

## Test Instructions
- Run docs:build on site without a `bodiless.docs.toc.js` and verify that default structure is correct.
- Create a `bodiless.docs.toc.js` at site level and verify that it controls docs structure
- Create a package containing a namespaced `docs.json` file, update toc to refer to it, and verify that structure is correct.


packages/bodiless-documentation/src/write.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Whether a file is copied or symlinked is determined by the `BODILESS_DOCS_COPYFILES` environment
+ * variable.
+ *
+ * @param docPath The absolute path of the source document.
+ * @param filePath The path of the target document relative to `process.cwd`
+ */
+const copyOrLinkFile = (docPath: string, filePath: string) => {
+  let relPath;
+  try {
+    if (process.env.BODILESS_DOCS_COPYFILES === '1') {
```
e2thex: Seems like copyfiles should be a prop and just default to the env variable

- [ ] - make copyfile a prop or decide not to

wodenx: do you mean it should be an argument to copyOrLInkFile() ?

e2thex: Yes  I think there should be a prop in copyOrLinkFile and the default can be to look at the env variable or better yet it just has a prop and there is a different function that uses the env bariable and passes it to copyOrLinkFile

wodenx: why?  do you anticipate there being a third alternative?

e2thex: I just think it is cleaner to not have env variable buried inside of functions.  It basically a side effect.  For example one should not have to test the env variable system to test this function

packages/bodiless-documentation/src/write.ts
```diff
@@ -0,0 +1,113 @@
...
+ * @param docPath The absolute path of the source document.
+ * @param filePath The path of the target document relative to `process.cwd`
+ */
+const copyFile: Copier = (docPath: string, filePath: string) => {
+  const relPath = path.relative(process.cwd(), docPath);
+  try {
+    return fs.copy(relPath, filePath, { overwrite: false });
+  } catch (error) {
+    console.warn('Error writing', filePath, relPath, error);
+    return Promise.resolve();
```
e2thex: Why are we not return rej?

packages/bodiless-documentation/src/write.ts
e2thex: I just think it is cleaner to not have env variable buried inside of functions.  It basically a side effect.  For example one should not have to test the env variable system to test this function

packages/bodiless-documentation/src/write.ts
dewen: Or do we need try/catch here, since fs.copy returns a Promise, error will be handled by caller.

If we do need console warning here, might use async/await to catch error here?

Issue #115, opened 2020-02-06
Alter the Main Menu in Preview Mode


hvanyo:
## User Story

As a Content Editor, I want to easily distinguish between Preview and Edit modes, so that it is clear what version I am working in. 
 * ~~Preview Mode should be seen as the default view~~ [UPD: 13 Feb, there is no agreement on this req. To be discussed with POBA team]
 * Preview Mode should only include the Docs, Edit and Last Commit icons
 ** All other icons will be hidden.
 ** The Last Commit icon will display below the edit icon
 * Current problem this solves is the ability to create a page was enabled even when in preview mode.  It should be disabled.

### Acceptance Criteria

GIVEN that a user is in Preview Mode
THEN they will see the Preview Admin Menu; Location Switcher, Docs, Edit and Last Commit
    AND the icons work as they currently do

WHEN the user selects the Edit icon
THEN they will see the Full Edit UI Admin Menu (current view of the menu)
    AND the icons work as they currently do

PR #132, opened 2020-02-12
feat(core): Alter the Main Menu in Preview Mode

VancheeZze:
## Changes

- New optional `isHidden` prop on all contextMenu buttons.
- Preview Mode now only includes the Docs, Edit, and Last Commit icons.

## Test Instructions

- Run `npm run start`
- In preview mode, we should only see Docs, Edit, and Last Commit icons.
- In edit mode, we should still see all the icons.

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/115


TatsianaMiraniuk:
The bug is fixed. Thanks!

Issue #85, opened 2020-01-28
Component Default Content


wodenx:
## User Story

- AS A Site Editor
- I WANT to receive components which have default content
- SO THAT I can quickly apply upstream updates

### Acceptance Criteria

AC1.
A developer can supply default content for any component (making it a "contentful" component.

AC2.
Default content can be supplied for a single component, or a tree of nested subcomponents.

AC3. 
The "contentful" component(s) will render their default content until the Editor overrides the content (edits the content) 

AC4.
A content editor has the ability to revert to the default content for subcomponent by subcomponent.

AC5.
Default content and assets can be packaged with a component, and updates received by upgrading the component package.  In such cases, any overridden content on the site will not be changed, but any subcomponent which has not been overridden will render the new default content.

AC6.
On the test site home page, the 2 touts are replaced with contentful components. The contentful components are available to use in flexbox areas. 

AC7. 
There are unit tests for this story. 

### Implementation Notes

- Wrap with a "withContent" HOC which provides a content node which supplies the default content when the store has no entry?
- Component children are deleted as well. 
- https://www.gatsbyjs.org/docs/importing-assets-into-files/

PR #144, opened 2020-02-17
feat(gatsby-theme-bodiless, bodiless-backend, bodiless-core, bodiless-richtext): Deleted Data Should Not be Retained

### LINKED ISSUE #65: Deleted Data Should Not be Retained
## User Story

As a Content Editor, 
I want default values for content in a new item (eg list element) when it is added after a previous item was deleted
So that I can start from a clean slate

As a site owner
I want stale data deleted from my site
So that my repo is clean.
 
### Acceptance Criteria

AC1: 
When I delete
 - a list item or its sublist item from a menu or list
- a component from flexbox
- a link or image from richt text editor
then all data owned by the deleted component or any of its children
are deleted from disk.

AC2. 
When you commit, the deleted Json files are also committed

### Implementation Notes

- Create a new `delete` method in content node which removes data from the store for that path and all sub-paths
- Create new route on backend which deletes the actual json files
- Consume the api in components (flexbox, list) which dynamically create children.
- We need to soft-delete items in GatsbyMobxStore (set a flag rather than removing from store) so that mechanisms for ensuring data integrity (preventing update with stale data per #14) continue to work.

beliayeu:
## Changes
* Created a new delete method in content node which removes data from the store for that path and all sub-paths
* Created new route on backend which deletes the actual json files
* Consume the api in components (flexbox, list) which dynamically create children.
* Soft-delete items in GatsbyMobxStore (set a flag rather than removing from store) so that mechanisms for ensuring data integrity (preventing update with stale data per #14) continue to work.

## Related Issues
Fixes #65

BREAKING CHANGE:
1. Submenu data model changed. The first reason is to make menu/submenu data model similar to list/sublist data models, so that menu/submenu can leverage api provided by list component. The second reason is to solve a submenu bug in which one json file stores data from multiple nodes. Particularly, submenu item stores toggle and a list sublist items. One, who has a submenu on a site, will have to either update existing submenu json files or recreate submenu. Example how to update submenu json files of existing site is demonstrated on the demo site. pr 41.
2. Accordion changes. Node is removed from SingleAccordionClean. The reason is to provide consumers more control over how the node is added to accordion. Particularly, this change was needed for burgermenu, which leverages accordions, so that  burgermenu can read submenu data from accurate node. One, who uses SingleAccordionClean to compose custom accordions, should inject node to the custom accordions.  Example can be found  in test-site/src/components/SingleAccordion/index.tsx. withNode has been added to asSingleAccordion.


packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
```diff
@@ -0,0 +1,213 @@
...
+  private deleteData() {
+    if (this.shouldSave()) {
+      this.store.client.deletePath(this.getResoucePath())
+        .then(() => this.updateState(ItemStateEvent.OnPostEnd));
+    } else {
+      this.updateState(ItemStateEvent.OnPostEnd);
+    }
+  }
+
+  private scheduleDataPost() {
```
wodenx: I think you can make this more functional by having a `scheduleRequest` method which takes a function (currently either `deleteData` or `postData`) as an argument.

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
```diff
@@ -0,0 +1,213 @@
...
+          this.setData(data);
+          this.updateState(event);
+        }
+        break;
+      default:
+        throw new Error('Invalid item event specified.');
+    }
+  }
+
+  delete() {
```
wodenx: so maybe just set isDeleted here (and clear it in setData) instead of doing so in the state machine.

packages/bodiless-components/src/List/model.ts
```diff
@@ -34,12 +38,13 @@ export const useItemsAccessors = () => {
...
   return (item: string) => {
-    if (getItems().length > 1) {
-      const items = getItems();
-      setItems(items.filter(item$ => item$ !== item));
-    } else if (unwrap) {
+    const items = getItems().filter(item$ => item$ !== item);
+    setItems(items);
+    deleteItem(item);
+    if (items.length === 0 && unwrap) {
+      deleteItem();
```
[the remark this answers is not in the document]
beliayeu: @wodenx, makes sense for me
tentative implementation - https://github.com/beliayeu/Bodiless-JS/compare/feat/retained-data...beliayeu:feat/retained-data-delete-list?expand=1

packages/bodiless-layouts/src/FlexboxGrid/model.ts
```diff
@@ -98,6 +102,10 @@ export function useFlexboxDataHandlers(): FlexboxDataHandlers {
       if (index >= getItems().length) return undefined;
       spliceItem(index, 1);
       const items = getItems();
+      deleteItem(uuid);
+      if (getItems().length === 0) {
+        deleteItem();
```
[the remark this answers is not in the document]
beliayeu: @wodenx, it means that once all items are removed from flexbox, we will end up with a json file containing empty array. but actually that is true for all other components. when text is removed from editable component, then we get json file with empty text, while probably it is better to have json file deleted. do we want to consider a mechanism to remove json file when component data is reverted to default value? or should it be as part of #85?

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  @action private setState(state: ItemState) {
+    this.state = state;
+  }
+
+  private updateState(event: ItemStateEvent) {
+    switch (event) {
+      case ItemStateEvent.UpdateFromBrowser:
+        this.isDeleted = false;
```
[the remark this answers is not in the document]
beliayeu: fixed in 951862392c61cc0d166af25f99fe9435a7215472

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
```diff
@@ -0,0 +1,213 @@
...
+        .then(() => this.updateState(ItemStateEvent.OnPostEnd));
+    } else {
+      this.updateState(ItemStateEvent.OnPostEnd);
+    }
+  }
+
+  private deleteData() {
+    if (this.shouldSave()) {
+      this.store.client.deletePath(this.getResoucePath())
+        .then(() => this.updateState(ItemStateEvent.OnPostEnd));
```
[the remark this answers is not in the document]
beliayeu: fixed in 951862392c61cc0d166af25f99fe9435a7215472

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
beliayeu: fixed in 951862392c61cc0d166af25f99fe9435a7215472

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
beliayeu: fixed in 951862392c61cc0d166af25f99fe9435a7215472

packages/bodiless-layouts/src/FlexboxGrid/model.ts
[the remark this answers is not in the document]
wodenx: Yes for #85 we'll add a "revert" button which will simply delete the json file for that component.

packages/bodiless-components/src/List/model.ts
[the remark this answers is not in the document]
wodenx: No actually I was thinking that the delete should be handled by the unwrap handler  in [`withToggleTo`](https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-components/src/Toggle.tsx#L39).

packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
wodenx: This isn't exactly what I meant - created a PR to illustrate a possible simpification.

packages/bodiless-components/src/List/model.ts
```diff
@@ -44,12 +44,19 @@ const useDeleteItem = ({ unwrap }: Pick<Props, 'unwrap'>) => {
...
-      deleteItem();
       unwrap();
     }
   };
 };
 
+/**
+ * Returns a method which can be used to delete a list
+ */
+const useDeleteList = () => {
```
wodenx: Can we rename this to `useDeleteSublist()` -- it took me a while to figure out what it was doing.

wodenx: And also - maybe don't pass in an item, since this is really not about deleting an item at all  so
```
const useDeleteSublist = () => {
  const { deleteItem }  = useItemsAccessors().
  return () => deleteItem('sublist');
}
```

wodenx: Maybe also rename `deleteItem()` to `deleteChild()` or `deleteSubnode` since it's used for more than just deleting items.

beliayeu: @wodenx, resolved in cef374193e6743e5c0f52f71e9024bf3fe86a69d and in 7b38eb60b2c58fc450338f832e1c6eb88af1b8fc
thanks for the feedback

TatsianaMiraniuk:
Hi @wodenx, @beliayeu ,

testing has been finished. Please see my notes below.

_Found defect (not reproduced in Master):_
**Steps to reproduce:**
1. open Home page or /rcmenu/
2. add a couple of sublist items in Main Menu item
3. delete all sublist items
4. add a new sublist item
**Actual Results:**
sublist is not added, not shown on hover; 'add a sublist item' button is hidden.
**Expected Results:**
no issues while adding the sublist items

Existing/Known issue for 'remove link' functionality:
URL is retained in Link form after removing a link (AESQ-2348)

**Questions:**
1. is the following behavior expected?
- List:  list2$<UUID>.json is created when UUID item gets a sublist. When all sublist items are deleted for that list item, this json is not deleted (but changed to {
  "on": false
}). 
- Flexbox: when a component is added into a flexbox area, the corresponding json is created. However when we delete that component (clean flexbox area completely), that json is not deleted, but gets updated.

2. Regarding AC2:
'When I delete
...
or image from rich text editor'
Currently we don't have a feature that would allow to add an image in RTE.
Does that AC also apply to Image component(single or inside Touts)?
In the current implementation, the images that were uploaded are retained. They are not deleted along with a component.  That needs to be discussed if that should be a requirement as tan image may be used in another component (since we are sharing the resources).

Thanks,
Tatsiana

wodenx:
@TatsianaMiraniuk For some reason I can't reply to your comment, so am posting here:

> Flexbox: when a component is added into a flexbox area, the corresponding json is created. However when we delete that component (clean flexbox area completely), that json is not deleted, but gets updated.

Yes this is expected.

> Does that AC also apply to Image component(single or inside Touts)?

No.  There is no way to delete a single image or image inside tout.

> In the current implementation, the images that were uploaded are retained. 

I agree this needs to be discussed.  Removing the images themselves is not in scope for this story, but we should decide if (and how) we want to tackle it.  cc @e2thex @hvanyo @schristodoulou 

> Existing/Known issue for 'remove link' functionality:
URL is retained in Link form after removing a link (AESQ-2348)

Yes. - discussed this with @beliayeu during development and agreed it was not in scope.

> List: list2${UUID}.json is created when UUID item gets a sublist. When all sublist items are deleted for that list item, this json is not deleted (but changed

This seems wrong - @beliayeu please investigate.

packages/bodiless-components/src/List/withSublist.tsx
```diff
@@ -26,25 +26,42 @@ import { useItemsMutators } from './model';
...
+  return {
+    unwrap: unwrap$,
+  };
+};
+
+/**
+ * Takes a component
+ * returns a new component with sublist node key and sublist deletion on unwrap
+ */
+export const asSublist = <T extends ListProps>(Sublist: ComponentType<T>) => (props: T) => {
```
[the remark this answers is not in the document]
wodenx: I see - you are using this HOC in withSubmenu - but I still think it might be cleaner just to export the hook, and use that directly in `withSubmenuToggle`, `withSublistToggle`, etc.

packages/bodiless-organisms/src/components/MainMenu/withSubmenu.tsx
```diff
@@ -53,7 +64,7 @@ const withSubmenuToggle = (Sublist: ComponentType<MenuSublistProps>) => (
 const withSubmenu = (Sublist: ComponentType<MenuSublistProps>) => (
   withDesign<ListDesignableComponents>({
     ItemMenuOptionsProvider: withToggleButton({ icon: 'playlist_add' }),
-    Item: withSubmenuToggle(Sublist),
+    Item: withSubmenuToggle(asSublist(Sublist)),
```
wodenx: See my comment in `withSublist.tsx` - rather than using this HOC maybe just use the hook (`useSublistUnwrap`) direclty in `withSubmenuToggle`

packages/bodiless-organisms/src/components/BurgerMenu/withBurgerSubmenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+} from '@bodiless/components';
+import { withDesign } from '@bodiless/fclasses';
+import {
+  useNode,
+  withNode,
+  withNodePath,
+} from '@bodiless/core';
+import { flow } from 'lodash';
+
+const withBurgerSubmenuToggle = (Sublist: ComponentType<ListProps>) => (
```
wodenx: It seems to me that this is doing the same thing as `withSubmenuToggle()`, only using `withNodePath `instead of creating a `NodeProvider`.  If so, then we should not duplicate -- and I believe using the `NodeProvider` is a better pattern as it avoids calling an HOC during a render cycle, which is a no-no.  But I still don't understand why we are providing a node with the same path as the parent node...

packages/bodiless-components/src/List/withSublist.tsx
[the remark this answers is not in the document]
beliayeu: the reason why I am abstracting this to an HOC is want to reuse the unwrap logic and sublist node key.  but probably I should make that sublist/submenu toggles are responsible for setting node keys for sublist elements

packages/bodiless-organisms/src/components/MainMenu/withSubmenu.tsx
```diff
@@ -17,27 +17,38 @@ import React, {
...
   (Item: ComponentType<PropsWithChildren<{}>> | string) => {
     const ItemWithSubmenu: FC<ListProps> = ({
-      children, unwrap, nodeKey, ...rest
-    }) => (
-      <Sublist title={children} unwrap={unwrap} {...rest} />
-    );
+      children, ...rest
+    }) => {
+      const { node } = useNode();
+      const children$ = <NodeProvider nodePath={node.path}>{children}</NodeProvider>;
```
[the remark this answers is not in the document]
beliayeu: @wodenx, the problem here is that the title uses different node keys in different toggle states. it happened after I added a separate nodeKey to submenus. in sublist state, sublist node key is added to the item nodepath.
we do not have this problem in list since the list renders sublist and sublist children at the same level, so sublist and its children inherit the same node path. but here we are integrating with rc-menu api and need to inject title as a submenu prop.

packages/bodiless-organisms/src/components/MainMenu/withSubmenu.tsx
beliayeu: makes sense, the only concern is that I will need to set node key for sublists in  withSubmenuToggle

packages/bodiless-organisms/src/components/BurgerMenu/withBurgerSubmenu.tsx
beliayeu: the different is the target element that we are wrapping by node provider
in submenu case, we are wrapping rc submenu title prop
in burgermenu case, we are wrapping accordion title design element
so I believe it makes sense to have withSubmenu and withBurgerSubmenu separate.

packages/bodiless-components/src/List/withSublist.tsx
```diff
@@ -26,25 +26,42 @@ import { useItemsMutators } from './model';
...
+  const { unwrap, nodeKey, ...rest } = props;
+  const { unwrap: unwrap$ } = useSublistUnwrap(unwrap);
```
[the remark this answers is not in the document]
beliayeu: fixed in 87aec2b86d6dabb07dbc774f39eb77084c4ffde2

packages/bodiless-core/src/ContentNode.ts
```diff
@@ -64,14 +65,18 @@ export class DefaultContentNode<D extends object> implements ContentNode<D> {
...
-    return new DefaultContentNode<E>(this.actions, this.getters, path);
+    return this.create<E>(path);
   }
 
   child<E extends object>(path: Path) {
     const paths = Array.isArray(path) ? path : [path];
     return this.peer<E>([...this.path, ...paths]);
   }
 
+  create<E extends object>(path: Path) {
```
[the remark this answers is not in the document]
beliayeu: fixed in 87aec2b86d6dabb07dbc774f39eb77084c4ffde2

packages/bodiless-organisms/src/components/MainMenu/asMainSubMenu.tsx
```diff
@@ -20,14 +20,20 @@ import {
...
  * with rc-menu <Submenu /> component
  */
 const asMainSubMenu = flow(
+  asRCMenuSublist,
   asMainMenu,
   withDesign({
-    Wrapper: replaceWith(stylable(SubMenu as ComponentClass)),
+    Wrapper: replaceWith(
+      flow(
+        stylable(SubMenu as ComponentClass),
```
wodenx: Why do we need this typecast here?

packages/bodiless-components/src/List/asBasicSublist.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { ComponentType as CT, PropsWithChildren } from 'react';
+import { FinalProps as ListProps } from './types';
+
+const asBasicSublist = (Sublist: CT<ListProps>) => (
```
wodenx: This name - along with asBurgerMenuSublist and asMenuSublist - is misleading.  It makes these sound like HOC's which apply to the Sublist component, when in fact they apply to the Item design element, and aren't really HOC's. 
 Also, I think that withSublistToggle is now too decomposed (really just a wrapper around withToggleTo - and the extra abstraction of useItemWithSublist is also confusing, since it actually returns both ItemWithSublist and ItemWithotuSublist).

I probably led us down a bad path in looking for more abstraction - and ended up making things more complex/confusing.  I don't think we can do better without a bigger refactor, and I don't want to take the time for it now -- but I do think we can make a couple of small improvements.

Let's get rid of `withSublistToggle` and put the withToggleTo inside asBasicSublist, etc.  Then at least it is a true HOC (it returns the toggle component, and we do away with that weird 'useItemWithSublist`).  We can then pass these to withSublist as
```
withSublist(asBasicSublist)(Sublist)(List)
withSublist(asMenuSublist)(Submenu)(Menu)
etc
```
or perhaps even better
```
withSublist(asBasicSublist(Sublist))(List)
```

so withSublist would look something like this:
```
const withSublist = withSublistToggle => withDesign({
  ItemMenuOptionsProvider: withToggleButton({ icon: 'playlist_add' }),
  Item: withSublistToggle,
});
```

That seems a fairly small change, and I think it would make things clearer by removing one layer of abstraction.  If you can do it in a few hours, I think it's worth it.

wodenx: Also pls be sure to update any docs - and add clear jsdocs for `asBasicSublsit` et al describing how they take a list component and return an HOC which can apply to the Item.

beliayeu: applied the fix in 00ec310d89874580c0f3c5e0e63131cbb591b953

TatsianaMiraniuk:
Hi @beliayeu,

testing has been completed. The issue found in the first round of testing (related to a sub-list in Main Menu) is not reproduced anymore.

However, there's a new defect on PDP page.

**Steps to reproduce:**
1. open /products/ page
2. create a new page
3. on the newly created PDP page, add some text 1st Accordion body
4. expand 2nd Accordion and other accordions
**Actual Results:**
the data entered in 1st accordion is shared among all accordions.
Screen Shot 2020-03-17 at 4 43 46 PM
Only one body.json is created and doesn't get accordion-[n]-prexif
Screen Shot 2020-03-17 at 8 03 03 PM
The page is freezing after clicking in 2nd and other accordion bodies.


Also, the following is still actual:
For List: list2${UUID}.json is created when UUID item gets a sublist. When all sublist items are deleted for that list item, this json is not deleted (but changed to {
"on": false
}). 


Thanks,
Tatsiana

beliayeu:
@TatsianaMiraniuk, thanks for testing
answers on https://github.com/johnsonandjohnson/Bodiless-JS/pull/144#issuecomment-600363901
1. pdp accordions. the bug caused by accordion changes.  the issue fixed by 33cbcfa4f55f79de74c80f07d925beaa571f9228. also, added information about accordion breaking changes to the pull request description
2. sublist item toggle is not deleted. I believe that is expected. The toggle is part of a list item, not part of the sublist. So as far as we delete sublist, but do not delete list item, the json file containing the toggle is not deleted. Considering this particular sublist item toggle bug/gap, I believe we should not fix it as part of this ticket as far as fixing just this sublist toggle problem will not bring us a lot of value (this problem adds one additional json file for each list when user adds and then removes sublist) and this problem is part of more global problem.  Once we remove/adjust component data and make it equal to default component data, then we can remove json file. We can consider solving this global problem in 85. @wodenx , what do you think?

wodenx:
@TatsianaMiraniuk @beliayeu 
re: 
> Also, the following is still actual:
For List: list2${UUID}.json is created when UUID item gets a sublist. When all sublist items are deleted for that list item, this json is not deleted (but changed to {
"on": false
}).

I agree with @beliayeu that this is expected behavior.  That node (list2$UUID.json} belongs to the item, not to the sublist, and should not be deleted when the sublist is deleted.  If I understand correctly, previously that node was used to store the list of items when the sublist was  toggled on, but that  has been remediated.  So the current expected (and correct) behavior is:
1. Create new list item in list2 - list2$UUID.json is not created
2. Add sublist - list2$UUID.json is created with value { isOn: true }
3. Delete sublist - list2$UUID.json is modified to value { isOn: false }
If that is the current behavior, then this is not a defect.
Thanks
C

TatsianaMiraniuk:
Thanks, @wodenx and @beliayeu !

The issue on PDP page has been checked - it works as expected now.

Approving the PP.

PR #148, opened 2020-02-18
fix(psh): Private npm registry for static site.

wodenx:


packages/bodiless-psh/resources/static/static.platform.sh
```diff
@@ -2,7 +2,18 @@ if [ -f static.platform.custom.sh ]; then
   source static.platform.custom.sh
 fi
 
+init_npmrc () {
+  echo "Creating .npmrc"
+  if [ $APP_NPM_REGISTRY ] && [ $APP_NPM_AUTH ] && [ $APP_NPM_NAMESPACE ]; then
+    echo "NPM Private registry for $APP_NPM_NAMESPACE is $APP_NPM_REGISTRY"
+    bash -c 'echo NPM Auth token is ${APP_NPM_AUTH:0:50}...'
+    echo "$APP_NPM_NAMESPACE:registry=https:${APP_NPM_REGISTRY}" >> .npmrc
```
dewen: Shall we use ">" instead of ">>" here? So it's not appending to existing .npmrc.

wodenx: fixed

Issue #124, opened 2020-02-11
 Migration tool enhancement to auto turn off Bodiless-Tailwind Theme.


hvanyo:
## User Story

As a migrator, I want bodiless tailwind styling automatically turned off so that I don't have to do this as a separate manual step and possibly forget to do this step and cause various theme issues. 

 

### Acceptance Criteria

AC1:  the migration tool before compilation, adds BODILESS_TAILWIND_THEME_ENABLED='0'  to env.site file.

AC2 : the migration tool has option to override this behavior

hvanyo:
Reopening as found issue that you can't re-enable tailwind with parameter in migration settings.json

{
  "url": "https://example.com/",
  "crawler": {
    "maxDepth": 1,
    "maxConcurrency": 1
  },
  "disableTailwind": false
}

PR #155, opened 2020-02-19
feat(migration_tool): Migration tool enhancement to auto turn off Bodiless-Tailwind Theme

VancheeZze:
## Changes
- New `setupEnvVars` step for the migration tool which currently disables `BODILESS_TAILWIND_THEME_ENABLED` before compilation.
- New optional `disableTailwind ` param for `migration-settings.json` which controles `BODILESS_TAILWIND_THEME_ENABLED`. Set to `true` by default.

## Test Instructions
- Run `npm run sites:clone-local starter`
- Run `npm run sites:update starter`
- Go to `cd sites/starter`
- Make sure `BODILESS_TAILWIND_THEME_ENABLED` is commented out in `.env.site` by default
- Run `npm run migrate`
- Make sure `BODILESS_TAILWIND_THEME_ENABLED` is now set to `0` and not commented out in `.env.site`.

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/124


VancheeZze:
> Can you add this disableTailwind an option to migration settings:
> https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-migration-tool/doc/Technical/MigrationTool.md
> 
> Give the reason we recommend it off for migration of sites. If the site is adding new bodiless JS components to migrated site while flattening, there may be a case you would want this enabled, but the risk is this styling can interfere with the migrated sites styling.

@hvanyo Documentation is now updated. Thank you!

PR #157, opened 2020-02-21
Chore: Example site package dependency update after v0.0.40.

dewen:
## Changes

- Package version updated
- Example site package-lock updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "bodiless-js",
   "description": "Bodiless-JS",
-  "version": "0.0.37",
+  "version": "0.0.40",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "bodiless-js",
   "description": "Bodiless-JS",
-  "version": "0.0.37",
+  "version": "0.0.40",
```
beliayeu: @dewen, I believe package-lock.json should have the same changes. we need to regenerate it

dewen: Good catch, regenerate the lock file might require additional regression test, will follow up.

Issue #166, opened 2020-02-25
Provide fallback on migration errors to generate plain html


dewen:
## User Story

- As user of migration tool, I want an option enabled by default, that changes the behavior of the migration tool so that if the migration encounters a page error, instead of skipping JSX page generation, it will push original html to the page file and report a message in the output that this was done, so that I can retrieve all pages from target site.

### Background/Description

There is an option, enabled by default, that changes the behavior of the migration tool so that if the migration encounters a page error, instead of skipping jsx page generation, it will push original html to the page file and report a message in the output that this was done.

If the option is disabled, then it will return to behavior where on page error, it skips generation of jsx files.

PR #175, opened 2020-02-27
feat(migration-tool): Provide fallback on migration errors to generate plain html.

dewen:
## Changes

- Creating a new template to wrap original body html in case of parsing error during scraping. 

## Test Instructions

- Create a test web page with syntax that could cause parsing issue. i.e semicolon outside of attribute value.
`<p style="padding-bottom:2px";>UN SOIN RESSOURCANT</p>`
OR try migrate page https://lepetitmarseillais.be/fr/soins-du-visage/lingettes-micellaires-demaquillantes-leau-florale-de-rose
- do `npm run migrate`

Expects:

1. No error displayed in console.
2. Warning message displayed in console indicating there's fallback action taken for the page.

## Related Issues

fixing #166


packages/bodiless-migration-tool/src/index.ts
```diff
@@ -82,7 +82,7 @@ class MigrationTool extends Command {
       trailingSlash: settings.trailingSlash || TrailingSlash.Add,
       transformers: settings.transformers || [],
       htmltojsx: true,
-      disableTailwind: settings.disableTailwind === undefined ? true : settings.disableTailwind,
```
VancheeZze: Why would we remove this line?

packages/bodiless-migration-tool/src/index.ts
```diff
@@ -82,7 +82,7 @@ class MigrationTool extends Command {
...
+      allowFailbackHtml: settings.allowFailbackHtml || 1,
```
VancheeZze: This might have a problem when you set `settings.allowFailbackHtml` to `false`, but here it will still be evaluated as `1` meaning there is no way to disable it.

packages/bodiless-migration-tool/src/index.ts
dewen: I was just finding it's a merge mistake, updated, thanks

packages/bodiless-migration-tool/src/page-creator.ts
```diff
@@ -151,9 +152,17 @@ export class PageCreator {
...
-      const targetComponentsPath = path.resolve(this.params.pagesDir, '../../components');
       const targetPageJsxPath = path.join(this.params.pagesDir, this.getPageFilePath(this.params.pageUrl, 'Page.jsx'));
+
+      try {
+        htmlToComponents.convert(this.params.bodyHtml);
+      } catch (error) {
+        if (this.params.allowFailbackHtml) {
+          this.writeContent(targetPageJsxPath, this.wrapComponentHtml(this.params.bodyHtml));
+          return;
+        }
```
VancheeZze: I think it would be nice to have a message for an operator explaining what has happened

packages/bodiless-migration-tool/conf/template_dangerous_html.jsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+class Page extends React.Component {
```
beliayeu: @dewen, do we preserve head elements (injected using helmet) when we switch to the fallback approach?

packages/bodiless-migration-tool/src/index.ts
```diff
@@ -83,6 +83,9 @@ class MigrationTool extends Command {
       transformers: settings.transformers || [],
       htmltojsx: true,
       disableTailwind: settings.disableTailwind === undefined ? true : settings.disableTailwind,
+      allowFailbackHtml: settings.allowFailbackHtml === undefined
```
beliayeu: @dewen, recommend renaming failback into fallback

packages/bodiless-migration-tool/src/page-creator.ts
```diff
@@ -151,9 +153,18 @@ export class PageCreator {
         .filter(item => item.scope === ComponentScope.Local)
         .map(item => `${item.component}.jsx`);
       const htmlToComponents = new HtmlToComponents(settings);
-      htmlToComponents.convert(this.params.bodyHtml);
-      const targetComponentsPath = path.resolve(this.params.pagesDir, '../../components');
       const targetPageJsxPath = path.join(this.params.pagesDir, this.getPageFilePath(this.params.pageUrl, 'Page.jsx'));
+
+      try {
+        htmlToComponents.convert(this.params.bodyHtml);
+      } catch (error) {
```
beliayeu: @dewen, I recommend considering a case when there is an invalid html in head. my understanding that we do not handle this case now

dewen: Approach will be addressed in follow up issue.

packages/bodiless-migration-tool/conf/template_dangerous_html.jsx
dewen: Will be addressed in a separate issue.

beliayeu: @dewen, I believe we need to address it as part of this ticket. otherwise, the migrated page with invalid html in body will not have head and will be broken for end-users. the separate issue is for automatic fallback (download page as static resource) when there is invalid html in head

VancheeZze:
@dewen Please update the title and description of this PR before merging.

VancheeZze:
@dewen Could you please also update documentation here: 

https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-migration-tool/doc/Technical/MigrationTool.md#configuration

I think it would be good to add `allowFallbackHtml` option there.

packages/bodiless-migration-tool/src/index.ts
```diff
@@ -83,6 +83,9 @@ class MigrationTool extends Command {
       transformers: settings.transformers || [],
       htmltojsx: true,
       disableTailwind: settings.disableTailwind === undefined ? true : settings.disableTailwind,
+      allowFallbackHtml: settings.allowFallbackHtml === undefined
+        ? true
+        : (settings.allowFallbackHtml === 1),
```
VancheeZze: I think there is a little logic error here. Even though it is enabled by default if I specify `allowFallbackHtml = true ` in `migration-settings.json` it will still be `false` in this step since `(true === 1)` will return `false`.

You can try this code here:  https://jsfiddle.net/
```js
const test = true;
alert(test === undefined ? true : (test === 1));
```

dewen:
> @dewen Could you please also update documentation here:
> 
> https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-migration-tool/doc/Technical/MigrationTool.md#configuration
> 
> I think it would be good to add `allowFallbackHtml` option there.

updated

PR #182, opened 2020-02-28
chore: Release v0.0.42

awm086:


asiarheyeu:
- @dewen created (or will create) PR to fix release notes for 41. It should be merged before this PR.
- @dewen will audit is it safe to point release tags 40-41 to the correct commit hashes. If no, another option is to publish a new tag.
- @awm086 please edit this PR to reflect only diff between 41 and 42
- Another option is to use Lerna publish with a custom branch 41. @VancheeZze can sure more info about it
- General note: release docs should be updated accordingly

Issue #77, opened 2020-01-23
Purge unused CSS


wodenx:
## User Story

- AS A site owner
- I WANT reduced css size on bodiless pages
- SO THAT there is a better load time experience for my visitors.

### Acceptance Criteria

AC1. 
When a page is loaded on the static site
Unnecessary css is not included.

- Remove Bodiless Admin CSS 
- Purge unused site level CSS

AC2.
Tailwind CSS is included in the site using the standard post CSS method as documented in https://www.gatsbyjs.org/docs/tailwind-css/

### Implementation/Testing Notes

https://www.gatsbyjs.org/packages/gatsby-plugin-purgecss/

- Should be tested locally and on Platform.SH
- Should run a full regression test; reviewing every page to make sure we haven't lost any styling.

hvanyo:
Will also fix https://github.com/johnsonandjohnson/Bodiless-JS/issues/42

PR #195, opened 2020-03-04
feat(gatsby-theme-bodiless, starter, test-site): purge unused css

beliayeu:
## Changes
* added tailwind theme compilation to gatsby-theme-bodiless using the standard postcss method
* added bodiless tailwind css and site css purging to gatsby-theme-bodiless
* css purging is configured only for static builds
* added ability to exclude particular css files from static build
* removed babel-plugin-tailwind-components from gatsby-theme-bodiless and from gatsby sites as far as we do not use css-in-js libraries
* configured css compilation in bodiless-ui using postcss
* removed previous build-css scripts

## Notes
* no changes made for the approach how compiled tailwind css is imported to a gatsby site. the compiled tailwind css is still imported in site's gatsby-browser.js since that the approach used in gatsby starters and described in gatsby docs
* css files imported using webpack css-loader are not purged, but site builders can exclude certain css resources from static build
* per measurements, no gatsby build performance degradation identified after enabling css purging
* null-loader is upgraded in gatsby-theme-bodiless as far as there is an issue with version available in gatsby https://github.com/gatsbyjs/gatsby/issues/22409

## Test Instructions
* css purging is enabled only for static builds, so in terms of css purging primarily the testing should be performed on static builds, recommend performing testing on test-site and/or demo sites
* css compilation approach has been changed. sanity testing is needed for development builds and migration flow builds.

## Related Issues
* fixes: #77


packages/gatsby-theme-bodiless/build-css.js
```diff
@@ -0,0 +1,87 @@
...
+      .map(pkg => path.resolve(bodilessPackagesBasePath, pkg))
+      .map(dir => path.resolve(dir, srcFilesGlobPattern));
+    buildCSSPlugins.push({
+      resolve: 'gatsby-plugin-purgecss',
+      options: {
+        tailwind: true,
+        develop: purgeWhileInGatsbyDevelop,
+        purgeOnly: [
+          ...siteTailwindCSS,
+          ...adminTailwindCSS,
```
hvanyo: @beliayeu  haven't quite worked through how all this works.
But we shouldn't have to process adminTailwind css file in any manner and make it smaller, should just exclude it altogether on static build.

beliayeu: fixed in 75e07ff6e8b50bf4ef224de88624ac5f6fd320a0

PR #203, opened 2020-03-09
fix(psh): Platform.sh site return 502 error instead of 404 page on non-existing pages.

VancheeZze:
## Changes
- Disable `passthru` for the static app on psh to display default 404 page

## Test Instructions
- On a static site open any non-existing page.
- You should see a default p.sh 404 page.


dewen:
Please add PR description. Thanks

PR #209, opened 2020-03-13
feat(layouts): Define Design for the Flexbox

VancheeZze:
## Changes
- Flexbox is now using Design System and `Wrapper` and `ComponentWrapper` are available to be styled by the Design System.

## Test Instructions
- The flexbox Elements on the Test Site have a padding-right of 20px and Flexbox itself has a margin-top and margin-bottom of 20px on Edit, Preview, and Static environments.

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/213


packages/bodiless-layouts/src/ComponentSelector/getFilteredComponents.tsx
```diff
@@ -26,11 +26,15 @@ const getFilteredComponents = (
   filters: Array<any>,
   searchString: string,
 ) => {
+  // Strip components with no categories such as Design components.
+  // eslint-disable-next-line
+  const flexboxComponents = components.filter(comp => comp.hasOwnProperty('categories'));
+
```
dewen: Why do we name it _flexboxComponents_ here?

VancheeZze: So when we use a Design System we also pass a "components" which are not included in a component picker. So here we are filtering Design components out and making sure only a Flexbox components are rendered inside component picker.

wodenx: @VancheeZze I'm not sure this filter is ideal.  We know from the start which componetns are actually related to flexbox design and which are intended to be available in the component picker.  Why not just omit the former by name?
```
const flexboxComponents = components.omit([ 'Wrapper', 'ComponentWrapper' ]);
```

VancheeZze: @wodenx I think `omit` may not work here since `components` is an array of functions due to how we currently add components to the Flexbox which looks like this:
Screen Shot 2020-03-18 at 1 43 45 PM

Where design components look like this:
Screen Shot 2020-03-18 at 1 43 02 PM

And flexbox components look like this:
Screen Shot 2020-03-18 at 1 43 16 PM

VancheeZze: Updated

VancheeZze:
> @VancheeZze We should remove the wrappers that got put on the components themselves.
> I know we did this for touts https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/examples/test-site/src/components/Tout/token.tsx#L55 and I product tout https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/examples/test-site/src/components/ProductTout/token.tsx#L30
> 
> Then fix flexbox's where these are used. Granted this will change margins slightly of example site as its different spacing but the touts didn't have correct spacing originally. IN doing so, then we have consistent usage that components don't pad themselves (unless they have a requirement to do so) and we do it flexbox component..

Updated

Issue #172, opened 2020-02-26
Provide default psh cache expiry and allow local override


hvanyo:
## User Story

AS A Site Builder
I WANT the caching setting enable by default with platform default of 6 hours
SO THAT I quickly enable CDN and not have to set this per site.

As A Site Builder
I want to be able to customize cache expiry for my site
So that I can meet business requirements.

As a Platform Maintainer
I want to be able to roll out changes to p.sh configuration globally
So that I can improve deployments for all

### Background/Description

Fastly caching requires  "expires" setting on .platform.app.yaml  

### Acceptance Criteria

AC1:
The .platform.app.yaml files which are written by bodiless-psh-init
Are created by merging specific whitelisted keys in the site’s (.platform.app.yaml & `edit` `*.platform.app.yaml`)
And global version from @bodiless/psh (packages/bodiless-psh/resources/static/static.platform.app.yaml & packages/bodiless-psh/resources/edit/edit.platform.app.yaml

AC2.
The merge should use a recursive algorithm
to ensure that individual keys of nested objects are preserved (eg lodash `merge`).

AC3.
There is a whitelist of keys maintained in @bodiless/psh
And only whitelisted keys will be merged.

AC4.
Non-whitelisted keys will be ignored,
And a warning message will be printed to the console.

AC5.
All files created by `bodiless-psh-init` are overwritten by default
Every time you run `bodiless-psh-init`.
(opposite of current behavior, in which you must specify an `--override` switch).

AC6
The default `static.platform.app.yaml` has a default cache expiration
For the `/` and `/___docs` added.

AC7.
Documentation is updated with overview of merge & update doc of any necessary changes.

### Implementation Notes

Needs to be tested on platform.sh and verify the expires value is set in the platform app YAML file. 
Any site can override the expires to meet their requirements.

### References

https://docs.platform.sh/configuration/app/web.html#locations

hvanyo:
@VancheeZze  implemented simpler version of AC1 which is 
The .platform.app.yaml files which are written by bodiless-psh-init
Are created by merging specific whitelisted keys in the site’s (static.platform.app.yaml)
And global version from @bodiless/psh (static.platform.app.yaml)

If we are all in agreement, lets update this AC1 to reflect this.

cc: @dewen  @wodenx  @schristodoulou

hvanyo:
@VancheeZze  says that AC7 has technical imlementation

[...]

PR #220, opened 2020-03-19
feat(psh): Provide default psh cache expiry and allow local override

VancheeZze:
## Changes
-  `.platform.app.yaml` files are now merged with the local versions taking precedence.
- Only keys from `/bodiless-psh/resources/.platform/platform.whitelist.yaml` will be merged with local `.platform.app.yaml` taking precedence.
- Other keys will default to `*.platform.app.yaml` in `@bodiless/psh`

#### Test if `expires` prop works:
- Deploy a site to p.sh using updated `@bodiless/psh` package.
- Check the `network` tab in the chrome console.
- You should see the `Cache-Control` and `Expires` headers enabled.

#### Test if `.platform.app.yaml` files merged right:
- Run `npm run sites:clone-local starter`
- Run `npm run sites:update starter`
- Go into `cd sites/starter`
- Add `"bodiless-psh-init": "bodiless-psh-init"` to the `scripts` section
- Run `npm run bodiless-psh-init`
- Check if the right files were generated. ( should copy the defaults if none exists )
- Try changing `.platform.app.yaml` file based on the `platform.whitelist.yaml` and running `npm run sites:update starter` and `npm run bodiless-psh-init`

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/172

BREAKING CHANGE: 
- Instead of just copying `*.platform.app.yaml` files it will merge existing local versions with a default version from `@bodiless/psh` with local versions taking precedence. Only keys from `/bodiless-psh/resources/.platform/platform.whitelist.yaml` will be merged


packages/bodiless-psh/src/bodiless-psh-init.ts
```diff
@@ -13,52 +13,110 @@
...
+        fs.writeFileSync(`${siteRootFolder}/.platform.app.yaml`, staticYaml);
+      }
+    },
+  );
+
+  copyfiles(
+    [`${pshFolder}/resources/edit/*`, `${siteRootFolder}/edit`],
+    { up: true, exclude: '**/*.platform.app.yaml' },
+    (err: any) => {
+      if (err) console.log('Error copying static app files', err);
```
dewen: "Error copying edit app files"

packages/bodiless-psh/src/bodiless-psh-init.ts
```diff
@@ -13,52 +13,110 @@
...
+/* eslint no-lonely-if: 0 */
+const fs = require('fs');
+const path = require('path');
+const jsYaml = require('js-yaml');
+const copyfiles = require('copyfiles');
+
+const siteRootFolder = process.env.INIT_CWD || path.resolve();
+const pshFolder = path.resolve(__dirname, '..');
+
+const readYaml = (folderPath: string, fileName: string) => {
```
dewen: Any reason we don't use just filePath as parameter, instead of path and file name?

packages/bodiless-psh/src/bodiless-psh-init.ts
```diff
@@ -13,52 +13,110 @@
...
+
+const generateEditYaml = (whitelist: object) => {
+  const defaultEditYaml = readYaml(`${pshFolder}/resources/edit/`, 'edit.platform.app.yaml');
+  const siteEditYaml = readYaml(`${siteRootFolder}/edit/`, '.platform.app.yaml');
+  const finalEditYaml = mergeByKey(defaultEditYaml, siteEditYaml, whitelist);
+
+  return jsYaml.safeDump(finalEditYaml);
+};
+
+const init = () => {
```
dewen: platform.custom.sh should be copied over if not exist, or this requirement has been changed?

VancheeZze: It is copied over.

packages/bodiless-psh/src/bodiless-psh-init.ts
```diff
@@ -13,52 +13,109 @@
...
+    [`${pshFolder}/resources/.platform/*`, `${siteRootFolder}/.platform`],
+    { up: true, exclude: '**/*.whitelist.yaml' },
+    (err: any) => {
+      if (err) console.log('Error copying app config files', err);
+    },
+  );
+
+  copyfiles(
+    [`${pshFolder}/resources/static/*`, siteRootFolder],
+    { up: true, exclude: ['**/*.platform.app.yaml', 'platform.custom.sh'] },
```
dewen, line 98: The original exclude should be for edit environment, we don't need tp exclude for static site.

VancheeZze: @dewen  Good catch! Updated. Thanks!

hvanyo:
docs is reset:
* Key 'expires' is not whitelisted. The default value of '6h' from '@bodiless/psh' will be used.

PR #231, opened 2020-03-24
feat(2543): bundle optimization

mereck:
Proposal for targeted optimization of production bundle


hvanyo:
Closing this and attached this patch of this file to the "Investigate ways of reducing bundle size"

PR #236, opened 2020-03-25
feat(example/test-site): Updated the test site to have the correct pa…

### LINKED ISSUE #235: Correct the flexbox padding on the test site
## User Story

As a tester I want to see the correct padding on the test site.

- AS A [who are you? eg Developer, Content Editor, etc.]
- I WANT [what is the feature you want?]
- SO THAT [what will you be able to do once the feature is implemented?]

### Acceptance Criteria

AC1.  All items in the flexbox have about 20px padding between themselves but the overall flexbox does not have padding between itself and its container.

### Implementation Notes

[edit 30-Mar] Gallery is out of scope

e2thex:
…dding in the flexbox example

## Test Instructions

Add items to any flexbox on the test site
observe that 
1. there is padding between each item
2. there is no padding around the flexbox

## Related Issues

Fixes #235


examples/test-site/src/components/Flexbox/token.tsx
```diff
@@ -4,8 +4,8 @@ import {
 } from '@bodiless/fclasses';
 
 const asFlexboxWithMargins = withDesign({
-  Wrapper: addClasses('my-5'),
-  ComponentWrapper: addClasses('pr-5'),
+  Wrapper: addClasses('-m-5'),
```
VancheeZze: The only concern I have regarding the negative margin is that the elements outside of the flexbox will be pushed slightly inside of the boundaries of a flexbox ( see screenshot ).
Screen Shot 2020-03-25 at 12 45 53 PM

I think this might be ok for the test site. 
cc: @hvanyo @e2thex @TatsianaMiraniuk

VancheeZze: There are 2 options to avoid this:
- First one is to add another wrapper component
- The second one is to avoid using a negative margin on a wrapper component and build a grid using `justify-content`. The drawback of this approach is that while spacing will be consistent between flexbox elements, it won't be possible to configure specific margins/paddings since `space-between` will distribute available empty space equally between flexbox items.

e2thex: Yes I think this is ok the dotted line is just identifying the block and I think it is ok if it overlays a bit around the item.  also I think in most places you would expect that there would be space between the flexbox and other items on the page. for example on the home page

TatsianaMiraniuk:
Hi @e2thex ,

please see the found issues:
1) there is a difference between Edit and Preview Mode/Static env.:
The items are wider in Preview Mode/Static Env
Screen Shot 2020-03-27 at 12 17 36 PM


2) the following pages with Flexbox didn't get 20px padding for their items:
/products/
/gallery-final/ 
Screen Shot 2020-03-27 at 1 28 20 PM

Thanks,
Tatsiana

e2thex:
@TatsianaMiraniuk 
1.  I think this first one is a different bug that is not created by my PR and we should gap/bug it.
2. I had the padding to the product page, but the gallery page is there for the tutorial and so I do not want to add it there.

hvanyo:
@e2thex  are we ok with pages that have empty flexboxes when they are served now overlap
i.e.  https://www.screencast.com/t/5mpAlirjs

TatsianaMiraniuk:
> @e2thex are we ok with pages that have empty flexboxes when they are served now overlap
> i.e. https://www.screencast.com/t/5mpAlirjs

@hvanyo, do you see that issue in the latest commit of this pr? It's not reproduced in my end, but I saw it in the very first commit.

hvanyo:
> > @e2thex are we ok with pages that have empty flexboxes when they are served now overlap
> > i.e. https://www.screencast.com/t/5mpAlirjs
> 
> @hvanyo, do you see that issue in the latest commit of this pr? It's not reproduced in my end, but I saw it in the very first commit.

Addressed when I did a fresh pull.

PR #249, opened 2020-03-31
fix(docs): Docs Homepage is lost on gh-pages

### LINKED ISSUE #237: Docs Homepage is lost on gh-pages
STR.
1. Visit https://johnsonandjohnson.github.io/Bodiless-JS/#/

ER.
Documentation homepage shows as expected (README from @bodiless/documentation)

AR.
Boilerplatge default docs homepage is shown.

Note - this seems to be an issue with the travis build - locally docs build properly.

konst8:
## Changes
- Fix paths (letter case typos) in docs.json files
- Add paths validation to prevent such issues in future (docs build ok on MacOS but fail on Linux)

## Related Issues
Fixes #237


.vscode/launch.json
```diff
@@ -108,5 +108,12 @@
             "trace": true, // verbose
```
VancheeZze, line 108: Are comments allowed in json files?

packages/bodiless-documentation/src/blDocsBuild.ts
```diff
@@ -69,6 +69,9 @@ const blDocsBuild = async () => {
     pathsList[0],
   );
 
+  // Validate the paths for letter-case typos.
+  validatePaths(getSimplePaths(paths));
```
dewen, line 73: Shall we catch the error for warning output here? Otherwise I don't see how we can get the msg if path invalid.

packages/bodiless-documentation/__tests__/tree.test.ts
```diff
@@ -183,3 +184,17 @@ describe('getTreeFromDir', () => {
     });
   });
 });
+describe('getSimplePaths', () => {
```
dewen, line 187: Can we add test case for validatePaths?

packages/bodiless-documentation/src/blDocsBuild.ts
konst8: The idea of this checking is to throw an error when the file path is case-sensitively incorrect. This is to make sure that we have the same build results on Mac and on Linux.
Node can work with case typos in paths on Mac, while on Linux it will fail.
https://stackoverflow.com/questions/13617863/can-i-force-node-js-require-to-be-case-sensitive

konst8: So if we just catch it (even with a warning), it will pass the build on Mac and then we will get misleading results on the server.

packages/bodiless-documentation/__tests__/tree.test.ts
konst8: Added.

packages/bodiless-documentation/src/blDocsBuild.ts
konst8: Added a test case that shows what is expected.

PR #252, opened 2020-04-02
test(organisms): better selectors for LinkToggle

### LINKED ISSUE #183: Automated Testing: smoke tests for Link Toggle
## User Story 

- AS A BodilessJS developer
- I WANT the Link Toggle component to be covered by the automated tests 
- SO THAT I can check that my code doesn't introduce any new bugs in Link Toggle functionality

### Acceptance Criteria
AC1: There are automated tests for the Link Toggle component 
AC2: The Link Toggle elements have ARIA unique identifiers

konst8:
## Changes
Improve selectors for LinkToggle component.

## Related Issues
Partially fixes #183


packages/bodiless-components/src/LinkToggle.tsx
```diff
@@ -24,6 +24,7 @@ const EmptyToggle = flowRight(
   ifEditable(
     withWrapOnSubmit,
     asBodilessLink(),
+    addProps({ 'aria-label': 'Link Toggle' }),
```
VancheeZze, line 27: I think that to make it work on both: `static` and `edit` envs we may want to move it to `withLinkToggle` declaration:

```js
const withLinkToggle = flowRight(
  addProps({ 'aria-label': 'Link Toggle' }),
  withNode,
  withToggleTo(EmptyToggle),
);
```

konst8: In that case we will get an error:
```
Invalid prop `aria-label` supplied to `React.Fragment`. 
React.Fragment can only have `key` and `children` props.
```
What if we wrap the component into Span for both edit and static? Updated the PR.

konst8: Per skype discussion -- we don't want to wrap the component in static due to possible regression. Reverting the last commit, so the aria-label prop will be presented only in edit mode.

TatsianaMiraniuk:
@konst8 , @VancheeZze   
1) after the changes, aria-label="Link Toggle" is available for Main Menu - is that expected?
2) the case when a link is applied to the text doesn't get aria-label:  //*[@class="my-3"]/a
Please see selectors spreadsheet.

Thanks,
Tatsiana

konst8:
> 1. after the changes, aria-label="Link Toggle" is available for Main Menu - is that expected?
Yes, this is is expected, since the LinkToggle component is used in the Main Menu:
https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-organisms/src/components/MainMenu/withEditableTitle.tsx#L25

PR #260, opened 2020-04-06
Migrate 404 page

### LINKED ISSUE #227: Migrate 404 page
## User Story

As a site flattener tool operator
I want to migrate 404 page
So that I can preserve existing site behavior

### Acceptance Criteria

AC1
Migration tool migrates special 404 page. By default it takes the page from /404 path

AC2
There is configuration in migration settings.json that allow to override the default path from AC1.

AC3
When migration tool hits a page different to the special 404 page and the server responds with 404 status
Then the page should not be migrated
Then the operator should be notified about non-existing page

konst8:
## Changes
- Make it possible to not migrate pages with "Page Not Found" (404) status
- Allow an operator to set a specific page that will be used as a template for default 404 page
- Allow disabling above functionality

## How to test
- try migrating some non-existing page
ER: the page should not be migrated, but should redirect to the default /404 page
- add the following option in migration-settings.json
```json
"page404Url": "https://example.com/some-default-404-page"
```
ER: The default 404 page should be created based on the assigned url.
- add the following option in migration-settings.json
```json
"isPage404Disabled": true
```
ER: non-existing pages are not skipped, they are migrated as normal pages.

## Related Issues
Closes #227


packages/bodiless-migration-tool/src/index.ts
```diff
@@ -87,8 +87,25 @@ class MigrationTool extends Command {
       allowFallbackHtml: settings.allowFallbackHtml === undefined
         ? true
         : (settings.allowFallbackHtml === true),
+      is404Source: false,
     };
     const flattener = new SiteFlattener(flattenerParams);
+    if (!settings.isPageNotFoundDisabled) {
```
beliayeu, line 93: I believe the issue with this approach is that the 404 page can be scraped twice. The first time is during page 404 flattening, the second time is during default flattening.

another issue is that the second round may overwrite the page we flattened during the first round (custom /404 is flattened the first time, and then it is overwritten due to the second round of flattening).

My recommendation is to explore opportunities to use crawler queue - https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-migration-tool/src/scraper.ts#L140-L144

packages/bodiless-migration-tool/src/scraper.ts
```diff
@@ -99,7 +99,12 @@ export class Scraper extends EE<Events> {
           }
           // decide if we get page or resource response
           if (successResult.isHtmlResponse) {
-            const { result } = successResult;
+            const { result, response } = successResult;
+            const isPath404 = response.url.replace(/.*\//, '') === '404';
```
beliayeu, line 103: it would be great to move this logic to a separate function so that make it more testable

packages/bodiless-migration-tool/src/scraper.ts
```diff
@@ -99,7 +99,12 @@ export class Scraper extends EE<Events> {
...
+            if (response.status.toString() === '404' && !isPath404) {
+              console.warn(`Page ${response.url} was not found`);
```
beliayeu, line 105: it would be great to enhance the message so that operator can understand better what happened with the page

packages/bodiless-migration-tool/src/scraper.ts
beliayeu, line 103: also, we should handle the case when response of custom path 404 page is 404

packages/bodiless-migration-tool/src/index.ts
konst8: Done.

packages/bodiless-migration-tool/src/scraper.ts
konst8: Replaced this logic with existing addTrailingSlashToUrl function.

packages/bodiless-migration-tool/src/scraper.ts
konst8: Done.

packages/bodiless-migration-tool/src/scraper.ts
konst8: My assumption was that we will still get some html even if the response status of the custom path is 404. So, the tool still will scrape something and in the end, we will get the same behaviour as on original site.

examples/starter/.env.site
```diff
@@ -36,7 +36,7 @@
 
 # The Site tailwind theme is enabled by default.
 # To disable it - uncomment the following line.
-# BODILESS_TAILWIND_THEME_ENABLED='0'
+BODILESS_TAILWIND_THEME_ENABLED='0'
```
beliayeu, line 39: I believe this should not be changed

beliayeu, line 39: I believe this should not be changed

packages/bodiless-migration-tool/src/page-creator.ts
```diff
@@ -206,9 +207,20 @@ export class PageCreator {
 
   private writeContent(targetPath: string, content: string) {
     // eslint-disable-next-line no-console
-    debug(`trying writing to ${targetPath}`);
-    ensureDirectoryExistence(targetPath);
-    fs.writeFileSync(targetPath, content);
+    const { pageUrl, page404Url } = this.params;
+    let finalTargetPath = targetPath;
+    if (pageUrl === page404Url) {
+      const fileName = finalTargetPath.replace(/.*\//, '');
```
beliayeu, line 213: what purpose does this replacement serve?

packages/bodiless-migration-tool/src/scraper.ts
```diff
@@ -50,6 +52,8 @@ interface Events {
 
 export interface ScraperParams {
   pageUrl: string,
+  page404Url: string,
+  isPage404Disabled: boolean,
```
beliayeu, line 56: I believe this logic should be moved to site-flattener. we need to pass a list of urls and its settings to the scraper only

packages/bodiless-migration-tool/src/page-creator.ts
```diff
@@ -206,9 +207,20 @@ export class PageCreator {
...
+      finalTargetPath = path.join(
```
beliayeu, line 214: move the logic to determine 404 page path from writeContent to getPageFilePath

packages/bodiless-migration-tool/src/page-creator.ts
```diff
@@ -206,9 +207,20 @@ export class PageCreator {
 
   private writeContent(targetPath: string, content: string) {
     // eslint-disable-next-line no-console
-    debug(`trying writing to ${targetPath}`);
-    ensureDirectoryExistence(targetPath);
-    fs.writeFileSync(targetPath, content);
+    const { pageUrl, page404Url } = this.params;
```
beliayeu, line 210: there is no logic to check if page is disabled or not here

packages/bodiless-migration-tool/src/index.ts
```diff
@@ -60,14 +61,22 @@ class MigrationTool extends Command {
...
+      ...userSettings,
+    };
     const flattenerParams: SiteFlattenerParams = {
       websiteUrl: settings.url,
       workDir: this.getWorkDir(),
       gitRepository: this.getGitRepo(),
       reservedPaths: ['404'],
       scraperParams: {
         pageUrl: settings.url,
+        page404Url: settings.page404Url,
```
beliayeu, line 78: * move page404Url and isPage404Disabled from scraperParams to a new flattenerParams property. for instance, 404Handler.

404Handler is responsible for:
* if page 404 scraping feature is enabled/disabled
* if a particular url is a a 404 url
* (optional) file path to 404 page - probably it is better to keep it in page-creator

konst8: Done.

packages/bodiless-migration-tool/src/page-creator.ts
konst8: This logic is covered in page404Hanlder now.

packages/bodiless-migration-tool/src/page-creator.ts
konst8: This file is not touched any more.

packages/bodiless-migration-tool/src/scraper.ts
konst8: Done.

packages/bodiless-migration-tool/src/page-creator.ts
konst8: Replaced with native URL methods and moved all related logic to a separate entity.

examples/starter/.env.site
konst8: Right. Reverted.

PR #261, opened 2020-04-06
fix(psh): Provide default psh cache expiry for /__docs

VancheeZze:
## Changes
- `expires` key was added for `/__docs` path.

#### Test if `expires` prop works:
- Deploy a site to p.sh using updated `@bodiless/psh` package.
- Open Docs
- Check the `network` tab in the chrome console.
- You should see the `Cache-Control` and `Expires` headers enabled.

## Related Issues
Fixes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/172

### DIFF
--- a/packages/bodiless-psh/src/bodiless-psh-init.ts
+++ b/packages/bodiless-psh/src/bodiless-psh-init.ts
@@ -56,7 +56,6 @@ const mergeByKey = (Source: any, Destination: any, Whitelist: any) => {
             console.log(`Key '${key}' is not whitelisted and not found in '@bodiless/psh' defaults. The value of ${Destination[key]} will be used.`);
             Object.assign(result, { [key]: Destination[key] });
           } else {
-            console.log(`Key '${key}' is not whitelisted. The default value of '${Source[key]}' from '@bodiless/psh' will be used.`);
             Object.assign(result, { [key]: Source[key] });
           }
         }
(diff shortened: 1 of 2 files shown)


packages/bodiless-psh/src/bodiless-psh-init.ts
```diff
@@ -56,7 +56,6 @@ const mergeByKey = (Source: any, Destination: any, Whitelist: any) => {
             console.log(`Key '${key}' is not whitelisted and not found in '@bodiless/psh' defaults. The value of ${Destination[key]} will be used.`);
             Object.assign(result, { [key]: Destination[key] });
           } else {
-            console.log(`Key '${key}' is not whitelisted. The default value of '${Source[key]}' from '@bodiless/psh' will be used.`);
```
hvanyo, line 59: @VancheeZze 
Will this produce a warning elsewhere if the key is "not whitelist" and default value != site value?

VancheeZze: Yep, Just added it.

packages/bodiless-psh/src/bodiless-psh-init.ts
```diff
@@ -56,7 +56,9 @@ const mergeByKey = (Source: any, Destination: any, Whitelist: any) => {
...
+            if (Source[key] !== Destination[key]) {
+              console.log(`Key '${key}' is not whitelisted and default value of '${Source[key]}' is different from site level value '${Destination[key]}'. Default value will be used. `);
```
dewen, line 60: Seems this doesn't work for array checking?

VancheeZze: It only happens to the array of simple items ( not pair/key arrays - see Supported YAML types here https://github.com/nodeca/js-yaml#supported-yaml-types ). I think this is expected.

PR #272, opened 2020-04-09
fix(migration-tool): site build failed due to invalid page content

beliayeu:
## Changes
* fixed the issue when there is a link to an image or other files (like pdf, doc, docx, etc) in href attribute of 'a' html element
* added content-type package to headless-chrome-crawler in order to detect type of response. particularly, to check whether the response is html or not.
* changed the logic how to detect whether the response is a file response or not in headless-chrome-crawler. all responses with content-type different to text/html are file responses now
* renamed 'success' event into 'pageReceived' event in migration-tool scraper so to avoid confusion
* updated headless-chrome-crawler types.
* updated unit-tests.

## Test Instructions
* migrate and build a website that has a page with a[href] that points to a file (image, pdf, doc, etc), or
*  start a test server on local machine
   - `git clone git@github.com:beliayeu/migration-tool-test-server.git`
   -  `npm i`
   -  `npm run start`
   - migrate and build the site served by the test server


packages/headless-chrome-crawler/lib/crawler.js
```diff
@@ -67,7 +67,8 @@ class Crawler {
       depth: this._depth,
       previousUrl: this._previousUrl,
       response: this._reduceResponse(response),
-      rawHtml: !isFile ? await response.text() : '',
+      responseText: !isFile ? await response.text() : '',
```
dewen, line 70: IsFile check seems limited to following content type, what about images and other binary types?

`  'application/octet-stream',
  'application/pdf',
  'application/msword',
  'application/zip',
  'application/vnd.ms-excel',
  'application/vnd.ms-powerpoint',
  'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
  'audio/mpeg',
`

beliayeu: the approach has changed in this pr. before this pr (the master version), the response is a file response when response content-type is one of 'application/octet-stream', 'application/pdf', etc. in this pr, all responses with content-type different to text/html are file responses.
so images and other binary types will be file responses, so migration tool will not create a page for them, but instead it will download them to static directory

Issue #265, opened 2020-04-07
Add Image Component


schristodoulou:
## User Story

- AS A Content Editor
- I WANT to add an Image to my page
- SO THAT I can follow the Brand's design

### Background/Description

- Users wanted an Image component and not just TACOs
- Users should not be confused by dimensions 
- We need to complete the UI abstraction API (at least iteration 1) and then use it in image component.

https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-components/src/Image.tsx#L115

### Acceptance Criteria

AC0: There is a landscape and square version of AC1 and AC2.  **updated landscape was previous horizontal 5/14/2020 with the following titles/descriptions:
Title: Square Image. Description: Adds a square image
Title: Square Linkable Image. Description: Adds a square linkable image
Title: Landscape Image. Description: Adds a landscape image
Title: Landscape Linkable Image. Description: Adds a landscape linkable image

AC1: There is an Image components (with all variations) in the Component Picker on examples/test_site.

AC2: The image component (and new default placeholders is used in the header image in the examples/starter_kit.  (uses the Landscape Image) 

AC3: Image default placeholders uses attached color image.  Image: https://user-images.githubusercontent.com/1383878/79777104-c1a2c800-8304-11ea-8972-2897071e52b1.png

AC4: All Images on examples/test_site & starter_kit use the default Image placeholder if they haven't overwritten the image (ie. TACO images have been replaced with the new image placeholder unless there is .json file assigning)   

AC5: Image upload supports JPG, PNG, GIF, SVG, WebP, APNG

AC6: Default placeholder can be overridden for an Image Component.

AC7: Remove use cases in the Examples/Test site where header image was overwritten (see use cases below) Image: https://user-images.githubusercontent.com/1383878/79788152-2ca8ca80-8316-11ea-9ba0-da6060dedc71.png

AC8: There is test page on examples/test_site with test identifiers to demonstrate Image Components in AC1 & AC2 with all supported image types and overriding default image.

AC9: 
All UI elements for the bodiless-components package
Are injectable via a ui prop on the Image component.
Image elements (ie. form) work as expected.

### Implementation Notes

Image Types:
https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types

Use cases where default image was used:
https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/examples/test-site/src/components/ProductListing/index.tsx#L41-L51

### References
[...]

PR #273, opened 2020-04-09
Feat(organisms) Carousel Image Only

hvanyo:
## Changes
* Replaced old Slider with Carousel
* Added Carousel that in edit mode can add/remove images.
* Added Example Carousel page in examples/test-site
* Cleaned up Organism Doc

## Test Instructions

- Edit PREVIEW: Visit http://localhost:8005/carousel/  and perform the following actions:
-- First Carousel rotates manually
-- Second Carousel rotates automatically
-- Carousel is responsive and acts as expected (similar behavior to Nuka carousel demo)
- Static: Visit  http://localhost:9000/carousel/ and perform the following actions:
-- First Carousel rotates manually
-- Second Carousel rotates automatically
-- Carousel is responsive and acts as expected (similar behavior to Nuka carousel demo)
- Edit Mode: Visit http://localhost:8005/carousel/  and perform the following actions:
-- First Carousel rotates manually
-- Second Carousel does NOT rotate automatically (because we need to edit form to work)
-- Carousel is responsive and acts as expected (similar behavior to Nuka carousel demo) with the exception that the mouse slide/dragging is not enabled (because we need to edit form to work)
 -- Can Add Carousel Images 
 -- Can Delete Carousel Images
 -- Can set Images as expected (image, alt)
 - Review Docs


beliayeu:
Testing feedback:
1. carousel paging items are tiny. is UI/UX included in this story?
2. when I add a new slide, I end up with the first slide, while to my mind it is better to end up with the new slide
3. default slide image does not fill container which leads to a lot of unusable white space
4. when I remove active slide, my slider is collapsed until I navigate to the next slide
5. when I upload an image with height greater than height of existing slides and navigate to one of existing slides, then I get unusable whitespace
6. when I delete some slides and then revert changes on the server (by doing git checkout), then my sliders are collapsed after hmr until I navigate to a slide

hvanyo:
> Testing feedback:
> 
> 1. carousel paging items are tiny. is UI/UX included in this story?

We were going to use default styling of nuka carousel library which are these tiny dots.  
It can be overwritten by props by site builder. 
 
> 2. when I add a new slide, I end up with the first slide, while to my mind it is better to end up with the new slide

To fix isn't trivial, I think as the new slide is added, I would have to adjust prop: slideIndex to be the new slide to show.   I don't think I will have time to address.  Enhancement to Carousel component in future?   

> 3. default slide image does not fill container which leads to a lot of unusable white space

This can be done with overriding the image prop but its kind of hack at this point.  there is future story coming that would let make this easier and default to short horizontal image. (https://github.com/johnsonandjohnson/Bodiless-JS/issues/265 -- AC6) 
Enhancement to Carousel component in future?   

> 4. when I remove active slide, my slider is collapsed until I navigate to the next slide

Not seeing this behavior. Example?

> 5. when I upload an image with height greater than height of existing slides and navigate to one of existing slides, then I get unusable whitespace

This is more the behavior of Nuka Carousel.  It takes on the height of the tallest image and doesn't resize.  I think the expectation here of MVP is a user would use all the same size images for the carousel.  We could expand on the requirements in future to resize/etc.

> 6. when I delete some slides and then revert changes on the server (by doing git checkout), then my sliders are collapsed after hmr until I navigate to a slide

think you are hitting this:
-> https://jira.jnj.com/browse/AESQ-648

> 7. page jump on static app on reload (my understanding that is a known issue)
You mean as page loads you see slider with no images and as images load the slider sizes to correct size.

beliayeu:
>> 4. when I remove active slide, my slider is collapsed until I navigate to the next slide

>Not seeing this behavior. Example?

made more experiments, the problem is reproduced on my side only when chrome dev tools is opened, so I believe we can descope this problem

packages/bodiless-organisms/src/components/Carousel.tsx
```diff
@@ -0,0 +1,147 @@
...
+  Wrapper: addProps({ 'data-carousel-element': 'wrapper' }),
+  Slider: withDesign({
+    Item: addProps({ 'data-carousel-element': 'slide' }),
+    Title: addProps({ 'data-carousel-element': 'slide-image' }),
+  }),
+});
+
+
+// Replace my Slider div with Slides List.
+// TO DO: This should move to Site Level because,
```
beliayeu, line 108: is the plan to do it as part of this pr or separate?

packages/bodiless-organisms/doc/Touts.md
```diff
@@ -0,0 +1,19 @@
...
+  - Provides a group on components commonly used as blocks on a site.
+- Properties
+  - none
+- Editable areas
+  - Title, Body, Image, CTA
+
+## Usage
+
+```js
+import { Tout, ImgTout, HTout } from '@bodiless/organisms';
```
beliayeu, line 13: per what I can see ImgTout, HTout do not exist in organisms now. Tout  is going to be removed in the default content component pr, so  I believe this doc can be removed

examples/test-site/src/data/pages/carousel/index.jsx
```diff
@@ -0,0 +1,41 @@
+/**
```
beliayeu, line 1: as far as the file contains types I would rename index.jsx into index.tsx

examples/test-site/src/components/Carousel/token.jsx
```diff
@@ -0,0 +1,26 @@
+/**
```
beliayeu, line 1: as far as the file does not contain jsx, I would rename it to token.js or even better to token.ts

hvanyo: Renaming to .ts

examples/test-site/src/data/pages/carousel/index.jsx
hvanyo: Renaming to tsx

packages/bodiless-organisms/doc/Touts.md
hvanyo: Removed

packages/bodiless-organisms/src/components/Carousel.tsx
hvanyo: It will be part of actual carousel story to help refine the slides.
I tried a few attempts to bring asEditableCarousel to site level and define the slide there but then it all started ending up at site level and not component.   So part of the actual carousel story will be making that slide component more exchangeable and definable at site level.  So I commented its not planned but don't plan on doing it here.

hvanyo:
@beliayeu  moved all issues to JIRA ticket as subtasks for when we discuss building out Carousel

PR #241, opened 2020-03-26
feat(organisms): "Filter by Tag" Component.

VancheeZze:
#### Purpose
   - Add a control that allows filtering content by group/tag with an ability to create a custom filter.
#### Properties
  - `suggestions` - a list of default suggestions (tags) - `[{id: '1', name: 'name'}, ...]`
#### Editable Areas
  - Filter component ( Filter Categories and Tags )

## Test Instructions
See https://github.com/johnsonandjohnson/Bodiless-JS/issues/223

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/223

PR #276, opened 2020-04-11
feat(Compoenet): "Filterable Metadata" Component

### LINKED ISSUE #222: Filterable item metadata
## User Story
As a content editor
I want to be able to add metadata (tags) to a component
So that I can allow an end user to hide it with a filter.

## Acceptance Criteria
AC1.
Any component can be wrapped (with one or more other components or HOCs) so that it provides an interface (form) for adding and removing tags. (see the wireframes "Groups" tag)

AC1a. There is a way to view All Groups (see example in wireframes "View All Groups" link). 

AC2.
The tags so added are stored as content (json files).

AC3.
The set of available tags can be constrained by the site builder when wrapping the component.

AC4
One can add a term that is not in the constrained list

AC5.
There is another wrapper which accepts a prop that is a function that takes tags
The component should call this function to pass its tags, if it returns false it should not show,
eg:
```
const myCallback = (tags: string[)) => logicThatDeterminesIfItShouldShow ? true : false;

<WrappedComponent shouldShow={myCallback} />
```

AC6.  
There exists a third wrapper which can wrap the above component
which causes it to show if it contains all of the tags passed as a prop, eg
```
<WrappedComponent tags={["foo", "bar"]} />
```
will render if the wrapped component has both foo and bar

AC7.
There is a page on the test site that includes a component wrapped (AC1-A6) for filtering. 
Along with allowing user to switch which tags should be used for filtering as well as multiple tag filters (foo + bar) 
(inputs to AC6).

## Implementation Notes:
- See wireframes https://jhdfdc.axshare.com/#g=1&p=product_lising&id=0sjqur 
- Unit Tests are also in scope. 
- The form autocompletes, so if the user enters an existing tag it will fill it in for them. 
- Validation requirements will be another story

awm086:
## Changes
- Add metadata (tags/groups) to a component that allows an end user to hide it with a filter.

## Test Instructions
- Visit /filter-item on the test site.
- See AC's at #222  

## Related Issues
Closes #222


packages/bodiless-components/src/Taggable/Item.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const { node } = useNode<Data>();
+  return {
+    getTags: () => node.data.tags,
+  };
+};
+
+export const editButtonOptions: EditButtonOptions<Props, Data> = {
+  icon: 'local_offer',
+  name: 'Add',
+  renderForm: ({ ui: formUi, props: props, formApi }) => {
```
wodenx, line 57: Let's be sure to consolidate with https://github.com/johnsonandjohnson/Bodiless-JS/pull/241/files#diff-83b14032dea19719c527a42f5fc0873bR28.

packages/bodiless-components/__tests__/Taggable.test.tsx
```diff
@@ -0,0 +1,152 @@
+import React from 'react';
```
awm086, line 1: move this too.

packages/bodiless-components/src/Taggable/Item.tsx
```diff
@@ -0,0 +1,156 @@
+// /**
```
awm086, line 1: delete this

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -0,0 +1,110 @@
...
+  };
+  const handleDelete = (i: number) => {
+    const newTags = tags.slice(0);
+    newTags.splice(i, 1);
+    setTags(newTags);
+    onChange(newTags);
+  };
+
+  return (
+    <>
```
awm086, line 88: @wodenx  adding ` <ComponentFormText type='hidden' name='tags' />` from informed is breaking submission of form.

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -0,0 +1,110 @@
...
+ */
+const useAccessors = () => {
+  const { node } = useNode<Data>();
+  return {
+    getTags: () => node.data.tags,
+  };
+};
+
+const ReactTagFieldBase = (props: Props) => {
+  const { getTags } = useAccessors();
```
wodenx, line 46: I don't think this field should be tied to the data model.

wodenx: What I mean here is that we should not be accessing the node to get the starting set of tags - they should be passed into this component as a prop.  We want this to be a pure informed form component which could be used on *any* informed form, in or out of bodiless (just like the text input field, the label, etc).  The only connection to bodiless is the UI prop which allows it to be styled.

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -0,0 +1,109 @@
...
+        suggestions={suggestions}
+        tags={tags}
+        placeholder={inputPlaceHolder}
+        noSuggestionsText={noSuggestionsText}
+        allowNew={allowNew}
+        addOnBlur={true}
+        handleDelete={handleDelete}
+        handleAddition={handleAddition}
+      />
+      <ComponentFormUnwrapButton
```
wodenx, line 98: I don't think this should be part of the component.  If you want, we could create a separate component (ShowTagListButton or whatever) which encapsualtes this.

packages/bodiless-core/src/withTagButton.tsx
```diff
@@ -0,0 +1,88 @@
+/**
```
wodenx, line 1: This should be in @bodiless/components not in @bodiless/core.

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -0,0 +1,109 @@
...
+        {...props}
+      />
+    </>
+  );
+};
+
+/*
+ * Hook to access tags.
+ */
+const useAccessors = () => {
```
wodenx, line 38: This should be moved into a separate module in @bodiless/components encapsulating the data model for tags (it will need to be shared with [the logic which consumes the tag list to show/hide items](https://github.com/wodenx/Bodiless-JS/commit/70c5f34b6e5b19b17d9cba4e08b6aa7dd45a9124#diff-f3dbe16cc82a836e94392dffc675f891R32-R37)

packages/bodiless-core/__tests__/hoc.test.tsx
```diff
@@ -1,10 +1,17 @@
 import React from 'react';
```
wodenx, line 1: Pls add tests for ifToggledOn and ifToggledOff

packages/bodiless-components/src/filterHOC.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * @param selectedTags
+ *  The selected tags to use.
+ */
+const useToggleByTags = <P extends object>({
+  selectedTags,
+}: P & ToggleByTagsProps) => {
+  const { getTags } = useTagsAccessors();
+  const itemTags = getTags();
+  const tags = itemTags.map(t => t.name);
+  return difference(selectedTags, tags).length === 0;
```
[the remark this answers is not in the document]
awm086: @wodenx I updated the structure but filtering but caused a regression that I'm unable to isolate. can you pull the branch and check.

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -39,13 +39,13 @@ export type ReactTagsFieldProps = {
...
   const formApi = useFormApi();
-  const currentTags = formApi.getValue<TagType[]>('tags');
+  const currentTags = formApi.getValue<TagType[]>('tags') || [];
   const { allowMultipleTags, ...rest } = props;
 
   const handleAddition = (tag: TagType) => {
     let tagToAdd = tag;
 
-    if (isEmpty(tagToAdd.id)) {
+    if (typeof tagToAdd.id !== 'number' && isEmpty(tagToAdd.id)) {
```
awm086, line 48: @VancheeZze I had to add this line because isEmpty(number) returns true.

examples/test-site/src/components/Filter/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const props = {
+    getSuggestions,
+    placeholder: 'Add or create',
+    formTitle: 'Groups',
+    seeAllText: 'See all groups',
+    formBodyText: 'Select from available tags:',
+    allowNew: true,
+    noSuggestionsText: 'No suggestions found',
+  };
+  const FilterableItemsList = getSuggestions().map(tag => (
```
wodenx, line 51: I don't think this list should be generated from the suggestions, since you can put any tags on any item.  I think this should just be. a list of taggable items (eg "Item 1", "Item 2", ...), otherwise it's confusing (eg, if I add the tag "Bar" to the item which prints "Foo")

examples/test-site/src/data/pages/filter-item/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+// @ts-ignore TS7016: Could not find a declaration file for module .. implicitly has an 'any' type
+import Layout from '../../../components/Layout';
+import TaggableFilterSelector from '../../../components/Filter';
+
+export default (props: any) => (
+  <Page {...props}>
+    <Layout>
+      <h1 className="text-3xl font-bold">Metadata (tags) Group Demo Page</h1>
+      <div className="my-3">
+        <TaggableFilterSelector />
```
wodenx, line 26: I don't think this is really a reusable component, so it should probably be defined at hte page level (in or alongside this file).

packages/bodiless-components/src/TagButton/withTagButton.tsx
```diff
@@ -51,7 +51,6 @@ export const tagButtonOptions:EditButtonOptions<TagButtonProps & HTMLProps<HTMLE
       seeAllText = 'See all tags',
       autofocus = false,
     } = componentProps;
-
```
wodenx, line 54: Is this change intentional?

packages/bodiless-components/src/Taggable/Item.tsx
```diff
@@ -0,0 +1,65 @@
+/**
```
wodenx, line 1: Please rename this file `asTaggableItem.tsx`

packages/bodiless-components/src/Taggable/Item.tsx
```diff
@@ -0,0 +1,65 @@
...
+// Note - the order is important. In particular:
+// - the node data handlers must be outermost
+// - anything relying on the context (activator, indicator) must be
+//   *after* `withEditButton()` as this establishes the context.
+// - withData must be *after* the data handlers are defiend.
+export const asTaggableItem = (nodeKey?: string) => flowRight(
+  withNodeKey(nodeKey),
+  withNode,
+  withNodeDataHandlers(emptyValue),
+  ifReadOnly(withoutProps(['setComponentData'])),
```
wodenx, line 48: Is there a reason we're removing this separately from the ones below?

packages/bodiless-components/src/filterHOC.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const ifTagsSelected = ifToggledOn(useToggleByTags);
+const ifTagsNotSelected = ifToggledOff(useToggleByTags);
+
+const withFilterByTags = flowRight(
+  ifTagsNotSelected(replaceWith(() => null)),
+  withoutProps(['selectedTags']),
+);
+
+export { ifTagsSelected, withFilterByTags, ifTagsNotSelected };
```
wodenx, line 56: Let's rename this file `withFilterByTags` and export that function as default.

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -19,7 +19,7 @@ import { isEmpty } from 'lodash';
 import ReactTags, { ReactTagsProps, Tag as ReactTagType } from 'react-tag-autocomplete';
 
 export type TagType = {
-  id: string,
+  id: string | number,
```
wodenx, line 22: Hm - do we really want to do this?  Why not just use string id's in the test case?

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -16,12 +16,7 @@ import React from 'react';
...
-import ReactTags, { ReactTagsProps, Tag as ReactTagType } from 'react-tag-autocomplete';
-
-export type TagType = {
-  id: string,
-  name: string,
-} | ReactTagType;
+import ReactTags, { ReactTagsProps, Tag as TagType } from 'react-tag-autocomplete';
 
 class Tag {
   readonly id: string = v4();
```
VancheeZze, line 22: Need to double-check if TagType.id can be a string 
or
Update this line to generate a unique number

packages/bodiless-core/src/components/ReactTagsField.tsx
```diff
@@ -40,7 +40,7 @@ const ReactTagsField = (props: ReactTagsFieldProps) => {
   const handleAddition = (tag: TagType) => {
     let tagToAdd = tag;
 
-    if (isEmpty(tagToAdd.id.toString())) {
+    if (tagToAdd.id && isEmpty(tagToAdd.id.toString())) {
```
VancheeZze, line 43: This seems wrong. If we will create a new tag ( which has no ID yet ) it will not go inside this `if (..)` statement.

When I create a new tag it equals to:
```js
tag = { name: 'NewTagName' };

if ( tag.id &&  isEmpty(tagToAdd.id.toString())) {...} // --> tag.id is always false for new tags.
```
I think
```js
if (!tag.id) {...}
```
should work in this case.

PR #279, opened 2020-04-15
Adding react tag field shared functionality

awm086:


packages/bodiless-components/src/TagButton/withTagButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </ComponentFormUnwrapButton>
+        </>
+      );
+    },
+
+    global: false,
+    local: true,
+
+};
+const withTagButton = () => withEditButton(tagButtonOptions);
```
awm086, line 80: @VancheeZze I removed the `options ` parameter and it's working fine not sure if we need it.

packages/bodiless-core/src/withEditButton.tsx
```diff
@@ -23,7 +23,10 @@ import { TMenuOptionGetter } from './Types/PageContextProviderTypes';
 
 export type FormBodyProps<D> = ContextMenuFormBodyProps<D> & {
   unwrap?: () => void;
+  // @todo: Make props a generic type.
+  props?: any;
```
wodenx, line 27: Can we rename this to "parentProps" or "formOwnerProps" or something to make it clear exactly which props they are?  Also add a jsdoc here describing them?

packages/bodiless-components/src/TagButton/withTagButton.tsx
wodenx: I think this is working bc you are getting the placeholder and no-suggestions text from props, whereas @VancheeZze was specifying them as options to the HOC.  I believe being able to bake them into the HOC is important (so you can easily set them at site level), but this can always be accomplished using another hoc which adds the props (like [`withPlaceholder`](https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-components/src/Editable.tsx#L80).  So I think this is fine.

packages/bodiless-core-ui/src/ReactTags.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { ReactTagsField, ReactTagsFieldProps } from '@bodiless/core';
+
+import './ReactTags.css';
+
+const ReactTags = (props: ReactTagsFieldProps) => {
```
wodenx, line 6: We should rename this to `ReactTagsField`

packages/bodiless-components/src/TagButton/withTagButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+            suggestions={suggestions}
+            placeholder={placeholder}
+            noSuggestionsText={noSuggestionsText}
+            minQueryLength={minQueryLength}
+            allowNew={allowNew}
+            allowMultipleTags={allowMultipleTags}
+            inputAttributes={inputAttributes}
+          />
+          <ComponentFormUnwrapButton type="button" onClick={displayListOfTags}>
+            See All Tags
```
wodenx, line 70: We need to be able to override this.  Probably should be another prop.

packages/bodiless-core-ui/src/ReactTags.tsx
awm086: @wodenx we already have ReactTagsField on line 21? should that be renames as well?

packages/bodiless-core/src/withEditButton.tsx
awm086: @wodenx @VancheeZze I ran into type issues when renaming.

Issue #285, opened 2020-04-20
PLP utilitizes Filtering system


hvanyo:
## User Story

- AS A site builder
- I WANT a filterable PLP
- SO THAT I can provide end-user a better UI experience in finding the product they need.

### Acceptance Criteria

AC1
On the examples/test_site there is filterable PLP page that uses the "Filter by Tag" component and product touts use the filterable metadata.

AC2
The PLP follows the UX design, https://jhdfdc.axshare.com/#g=1&p=product_lising&id=0sjqur. (Note the data in the prototype is the confirmed language to use as well.) 

AC3. 
The implementation should meet the the responsive behavior for All devices. 
- Tablet is mostly the same as Mobile. The width of the Filter button is full page in mobile where as in Tablet it is a 3rd. 
 
AC4.
There is a Product B PDP on the Test site.

AC5.
There are Product Touts on the PLP that links to the Product A & B PDPs. 

### Implementation Notes

Breakpoints from Tailwind -> https://tailwindcss.com/docs/responsive-design

/* Small (sm) */ (mobile)
@media (min-width: 640px) { /* ... */ }

/* Medium (md) */. (Tablet)
@media (min-width: 768px) { /* ... */ }

/* Large (lg) */. (Desktop)
@media (min-width: 1024px) { /* ... */ }

/* Extra Large (xl) */
@media (min-width: 1280px) { /* ... */ }

### References

- https://github.com/johnsonandjohnson/Bodiless-JS/issues/222
- https://github.com/johnsonandjohnson/Bodiless-JS/issues/223

PR #291, opened 2020-04-22
feat(components): PLP Utilizes Filtering System

VancheeZze:
## Changes
- FilterByGroup is now used on the PLP for filtering product touts.

## Test Instructions
See: https://github.com/johnsonandjohnson/Bodiless-JS/issues/285

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/285


packages/bodiless-components/src/List/ListAccordionTitle.tsx
```diff
@@ -35,15 +39,17 @@ const ListAccordionTitleBase: FC<ListAccordionTitleProps> = ({
...
   setExpanded,
+  alwaysExpanded,
   children,
   ...rest
 }) => {
   const { Wrapper, Icon } = components;
 
   return (
-    <Wrapper onClick={() => setExpanded(!expanded)} {...rest}>
+    <Wrapper onClick={() => setExpanded(alwaysExpanded ? true : !expanded)} {...rest}>
```
wodenx, line 49: Why not just remove the click handler here?

packages/bodiless-organisms/src/components/FilterByGroup/Filter.tsx
```diff
@@ -187,7 +187,7 @@ class FilterBase extends React.PureComponent {
     const { TagList, CategoryList } = components;
 
     this.RestProps = rest;
-    this.Filter = withBasicSublist(TagList)(CategoryList);
+    this.Filter = withAccordionSublist(TagList)(CategoryList);
```
wodenx, line 190: what about
```
this.Filter = flow(
  ifDesktop(withBasicSublist(...)(...)),
  ifNotDesktop(withAccordionSublist(...)(...)),
```

packages/bodiless-components/src/List/ListAccordionTitle.tsx
```diff
@@ -35,15 +39,17 @@ const ListAccordionTitleBase: FC<ListAccordionTitleProps> = ({
   components,
   expanded,
   setExpanded,
+  alwaysExpanded,
```
wodenx, line 42: Rather than adding a prop here, can we use the design api to swap out the accordion title for one that doesn't have this functionality?  Or even to swap out the whole sublist for one which isn't an accordion?

In other words, an accordion which is always expanded is no longer an accordion, so instead of adding stuff to make a version of an accordion which is always expanded, shouldn't we go the other way around.  Start with a normal list, and add stuff to make it an accordion when not on desktop.

What about something like,

Ideally we'd have an end api like
```
asResponsiveFilter = ifNotDesktop(
  asAccordionFilter
);
```
and
```
asAccordionFilter = withDesign(
  TagList: asAccordionSublist,
);
```
But this may not be easy/possible given the current state of withBasicSublist.

What about something like https://github.com/johnsonandjohnson/Bodiless-JS/pull/291/files#r417945400?

The point is that, in general, instead of building more complex components which can be configured to behave in different ways (ie adding props), we should be trying to build smaller, simple components and finding ways to compose the added functionality onto them or toggle between them.  See [this guidnig principle](https://johnsonandjohnson.github.io/Bodiless-JS/#/About/CorePrinciples?id=_3-favor-selection-over-configuration)

packages/bodiless-components/src/List/ListAccordionTitle.tsx
```diff
@@ -57,15 +63,23 @@ const ListAccordionTitleClean = flow(
...
+  ifDesktop(
+    addProps({ alwaysExpanded: true }),
+  ),
+);
+
 const ListAccordionTitle = flow(
   withDesign({
-    Wrapper: addClasses('flex justify-between w-full'),
-    Icon: addClasses('material-icons cursor-pointer select-none'),
+    Wrapper: addClasses('flex justify-between items-center'),
```
wodenx, line 74: I don't think we should be adding tailwind styling here.  We're trying to keep bodiless-components free of tailwind, so that the basic functionality can be used even if you choose a different css paradigm.

packages/bodiless-core/src/PageEditContext/index.tsx
```diff
@@ -42,6 +43,22 @@ export const reduceRecursively = <T extends any>(
     ? reduceRecursively(newAccumulator, callback, context.parent)
     : newAccumulator;
 };
+
+export const getSize = (size: number) => {
```
wodenx, line 47: I'm a little concerned about adding this functionality directly to PageEditContext in bodiless-core.  I'm wondering if we should create a new package (bodiless-responsive?) and/or put this in bodiless-components.  We could create a ResponsiveContextProvider which either extends the PageEditContext with the new functionality (or just provides its own context).

packages/bodiless-core/src/PageEditContext/index.tsx
```diff
@@ -88,6 +105,14 @@ export class PageEditStore implements PageEditStoreInterface {
     timeoutId: 0,
   };
 
+  @observable innerWidth: number = window.innerWidth;
+
+  constructor() {
```
wodenx, line 110: Should this be done in the constructor here? Or in componentDidMount somewhere?  Also - we'll need to check to be sure that window is defined.

packages/bodiless-core/src/PageEditContext/index.tsx
```diff
@@ -42,6 +43,22 @@ export const reduceRecursively = <T extends any>(
     ? reduceRecursively(newAccumulator, callback, context.parent)
     : newAccumulator;
 };
+
+export const getSize = (size: number) => {
+  type Sizes = {
+    [size: string]: number;
+  };
+
+  const sizes: Sizes = {
```
wodenx, line 52: This should definitely be a prop - we don't want to hardcode breakpoints here.

packages/bodiless-organisms/src/components/FilterByGroup/Filter.tsx
wodenx: We could even make this injectible eg add a `withSublist` prop - 
```
const { components, withSublist = withBasicSublist, ...restProps } = props;
...
this.Filter = withSublist(TagList)(CategoryList);
```

then maybe you coudl do
```
asResponsiveFilter = ifNotDesktop(
  addProp({ withSublist: withAccordionSublist, }),
);
```
This might be overkill - depends on how configurable we want our Filter to be.

packages/bodiless-organisms/src/components/FilterByGroup/FilterByGroupClean.tsx
```diff
@@ -33,23 +38,37 @@ const FilterByGroupBase: FC<FilterByGroupProps> = ({
...
 
+  const isExpanded = deviceSize.size === 'lg' || expanded;
+
   return (
     <Wrapper {...rest}>
       <FilterWrapper>
-        <ResetButton onClick={() => setSelectedTag()}>{resetButtonText}</ResetButton>
-        <Filter />
+        <FilterHeader>
+          <ListAccordionTitle expanded={isExpanded} setExpanded={setExpanded}>
```
wodenx, line 64: Again - is there a way here to use the design api to add the accordion functionality when not on desktop?

packages/bodiless-organisms/src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+      </Span>
+    </Div>
+  );
+};
+
+const asAccordionBody = <P extends Object>(Component: ComponentType<P>) => (props: P) => {
+  const { expanded } = useAccordionContext();
+
+  return (
+    <Div className={expanded ? 'block' : 'hidden'}>
```
wodenx, line 74: Do we need to wrap this in a div?  Or can we just add a 'hidden' class to the component when not expanded  - or, even better, just not render the component when not expanded...

packages/bodiless-components/src/withFilterByTags.tsx
```diff
@@ -36,6 +36,12 @@ const useToggleByTags = <P extends object>({
 }: P & ToggleByTagsProps) => {
   const { getTags } = useTagsAccessors();
   const tags = getTags();
+
+  // Show all items if there is no selected tag.
```
wodenx, line 40: Is this necessary? I thought this was the default behavior of difference -- and it was working before, no?

examples/test-site/src/components/ProductTout/index.tsx
```diff
@@ -26,6 +29,20 @@ import {
 } from '../Elements.token';
 import asProductToutDefaultStyle from './token';
 
+const withTagButtonProps = <P extends object>(Component: CT<P>) => (props: P) => {
```
wodenx, line 32: Why is this site-level?  Should be part of the filte rin bodiless-organisms, i think...

examples/test-site/src/components/ProductTout/index.tsx
```diff
@@ -42,3 +59,13 @@ export const ProductTout = flow(
   asProductTout,
   asProductToutDefaultStyle,
 )(ProductClean);
+
+export const asFilterableProductTout = flow(
```
wodenx, line 63: This shoudl be part of organisms too - asFilterableByGroup or something - it is not specific to products, this page, or the site.

examples/test-site/src/components/ProductTout/index.tsx
```diff
@@ -42,3 +59,13 @@ export const ProductTout = flow(
...
+  withFilterByTags,
+  asTaggableItem(),
+  withTagButtonProps,
+);
+
+export const FilterableProductTout = flow(
```
wodenx, line 69: We can probably just apply this when creating the variations for the plp flowcontainer

examples/test-site/src/data/pages/filter-page/index.tsx
```diff
@@ -23,6 +23,16 @@ export default (props: any) => (
   <Page {...props}>
     <Layout>
       <h1 className="text-3xl font-bold">Filter Demo Page</h1>
+      <p className="p-5">
```
wodenx, line 26: What's this for?

packages/bodiless-components/src/List/ListAccordionTitle.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ListAccordionComponents,
+  ListAccordionTitleProps,
+} from './types';
+
+const listAccordionComponentsStart:ListAccordionComponents = {
+  Wrapper: Div,
+  Icon: Span,
+};
+
+const ListAccordionTitleBase: FC<ListAccordionTitleProps> = ({
```
wodenx, line 34: There's duplication here with [`asAccordionTitle`](https://github.com/johnsonandjohnson/Bodiless-JS/pull/291/files#diff-c4faa3e7096414edee57129a0bfa07adR53-R68).  Can we fix that?

packages/bodiless-core/src/withEditButton.tsx
```diff
@@ -63,7 +63,7 @@ export const createMenuOptionHook = <P extends object, D extends object>({
     const submitValues = (values: D) => {
       setComponentData(values);
       Object.assign(componentData, values);
-      if (onSubmit) onSubmit();
+      if (onSubmit) onSubmit(Object.assign(componentData, values));
```
wodenx, line 66: Is this necessary?  Where are we using the values in onSubmit ?

examples/test-site/src/components/ProductListing/ProductListingFlowContainer.tsx
```diff
@@ -15,18 +15,19 @@ import React from 'react';
...
 import { FlowContainer } from '@bodiless/layouts-ui';
 import withProductVariations from './withProductVariations';
-import { asFlowContainerWithMargins } from '../FlowContainer/token';
+import { asFlowContainerWithMargins, asFlowContainerFullWidth } from '../FlowContainer/token';
 
 const withProductStrictSnapSize = Component => props => (
   <Component
     {...props}
-    snapData={() => ({ className: 'w-full lg:w-1/4' })}
+    snapData={() => ({ className: 'w-1/3' })}
```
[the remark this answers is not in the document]
VancheeZze: Before we had Filter by group each tout was 25% of the width so there were 4 touts in a row. When we added Filter by group on the page, the filter itself takes 1/4 so that all touts take the rest 1/3 each.

examples/test-site/src/data/pages/filter-page/index.tsx
VancheeZze: I just added here to have some content above FilterByGroup. On mobile admin menu covers most of the filter components. Added it here just to push the filter down.

packages/bodiless-components/src/withFilterByTags.tsx
VancheeZze: Without this, on initial render, we see no products displayed. Also when display all products clicked it filters out all product touts.

packages/bodiless-organisms/src/components/FilterByGroup/FilterByGroupClean.tsx
```diff
@@ -33,11 +39,14 @@ const FilterByGroupBase: FC<FilterByGroupProps> = ({
   components,
   children,
   resetButtonText = 'Reset',
+  filterTitle = 'Filter',
```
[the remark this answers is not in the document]
VancheeZze: I think `filterTitle` is designable so it actually can be made Editable through the site:

```js
const withEditableTitle = withDesign({
  FilterTitle: asEditable('filter_title', 'Filter Title'),
});
```

examples/test-site/src/components/FilterByGroup/index.tsx
```diff
@@ -45,15 +45,24 @@ const withCategoryList = withDesign({
   Title: addClasses('font-bold'),
 });
 
+const asResponsiveFilterByGroup = withDesign({
```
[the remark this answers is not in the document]
VancheeZze: Updated to use `token.tsx`

packages/bodiless-components/src/Taggable/asTaggableItem.tsx
```diff
@@ -12,7 +12,7 @@
  * limitations under the License.
  */
 
-import { HTMLProps } from 'react';
+import React, { HTMLProps, ComponentType as CT } from 'react';
```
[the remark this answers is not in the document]
VancheeZze: Renamed to `ComponentType `.

packages/bodiless-components/src/Taggable/asTaggableItem.tsx
```diff
@@ -23,15 +23,36 @@ import {
...
 // @Todo: Determine if this type is necessary?
 type Props = HTMLProps<HTMLElement>;
 
-const emptyValue = {
-  tags: '',
+type WithRegisterSuggestionsType = {
+  registerSuggestions: (tags: TagType[]) => void,
+};
+
+type ComponentTagData = {
```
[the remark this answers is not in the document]
VancheeZze: You were right, updated.

examples/test-site/src/components/ProductTout/index.tsx
VancheeZze: Moved to `@bodiless/organisms`

examples/test-site/src/components/ProductTout/index.tsx
VancheeZze: Done

packages/bodiless-components/src/List/ListAccordionTitle.tsx
VancheeZze: Fixed

packages/bodiless-core/src/withEditButton.tsx
VancheeZze: It is no longer needed, removed.

packages/bodiless-organisms/src/components/FilterByGroup/Filter.tsx
```diff
@@ -74,6 +77,19 @@ const withTagButtonProps = <P extends object>(Component: CT<P>) => (props: P) =>
   return <Component {...props} {...tagButtonProps} />;
 };
 
+const withOnTagDelete = <P extends object>(Component: CT<P>) => (props: P) => {
```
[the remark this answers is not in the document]
VancheeZze: Updated

packages/bodiless-organisms/src/components/FilterByGroup/Filter.tsx
```diff
@@ -177,7 +194,7 @@ const TestFilterComponentsStart: FilterComponents = {
 };
 
 class FilterBase extends React.PureComponent {
-  Filter = Div;
+  Filter: FC = Div;
```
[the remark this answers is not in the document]
VancheeZze: Updated

packages/gatsby-theme-bodiless/src/dist/Page.tsx
```diff
@@ -48,21 +49,27 @@ const getUI = (ui: UI = {}): FinalUI => ({ ...defaultUI, ...ui });
 
 const Page: FC<Props> = observer(({ children, ui, ...rest }) => {
   const { PageEditor: Editor, ContextWrapper: Wrapper } = getUI(ui);
+  const breakpoints = {
```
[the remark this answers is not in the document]
VancheeZze: Updated. I created a new Page component on the test site which takes the original Page and adds a `breakpoints ` prop to it using `assProps(...)`. I only used this updated Page component on Filter Page and PLP Page re reduce testing scope ( since Page component is used on every page and currently imported from `@gatsby-theme-bodiless` ). Lets discuss if we want to update all pages.

packages/bodiless-organisms/src/components/Accordion/index.tsx
VancheeZze: Updated. Not rendering component if not expanded.

examples/test-site/src/components/ProductTout/index.tsx
VancheeZze: Moved to `@bodiless/organisms`.

packages/bodiless-components/src/Taggable/asTaggableItem.tsx
```diff
@@ -23,15 +23,36 @@ import {
...
+  tags?: TagType[],
+};
+
+const emptyValue:ComponentTagData = {
+  tags: [],
+};
+
+const useRegisterTags = <P extends WithRegisterSuggestionsType>(Component: CT<P>) => (props: P) => {
```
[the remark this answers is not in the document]
VancheeZze: Updated. Renamed to `withRegisteredTags` and moved to the `withTagButton `.

examples/test-site/src/templates/product_listing.jsx
```diff
@@ -14,9 +14,10 @@
 
 import React from 'react';
 import { graphql } from 'gatsby';
-import { Page } from '@bodiless/gatsby-theme-bodiless';
+// import { Page } from '@bodiless/gatsby-theme-bodiless';
```
wodenx, line 17: Can we remove this comment?

packages/bodiless-components/src/TagButton/withTagButton.tsx
```diff
@@ -12,22 +12,21 @@
...
-
-type Data = {
-  tags: TagType[],
-};
+import { TagButtonProps, TagsNodeType, WithRegisterSuggestionsType } from './types';
 
 // Options used to create an edit button.
-export const tagButtonOptions:EditButtonOptions<TagButtonProps & HTMLProps<HTMLElement>, Data> = {
+// eslint-disable-next-line max-len
+export const tagButtonOptions:EditButtonOptions<TagButtonProps & HTMLProps<HTMLElement>, TagsNodeType> = {
```
wodenx, line 29: Can we fix this linting error rather than disabling the rule?

packages/bodiless-organisms/src/components/FilterByGroup/index.tsx
```diff
@@ -14,13 +14,15 @@
...
 
 export {
   FilterByGroupClean,
   TestableFilterByGroup,
   asTestableFilterByGroup,
+  asFilterableByGroup,
   withFBGSuggestions,
   useFilterByGroupContext,
-  withRegisterSuggestions,
+  withSuggestionProps,
```
wodenx: Per discussion in skype - let's rename to `withTagProps`

packages/bodiless-components/src/TagButton/withTagButton.tsx
VancheeZze: Updated, fixed the warning

examples/test-site/src/templates/product_listing.jsx
VancheeZze: Removed

packages/bodiless-organisms/src/components/FilterByGroup/index.tsx
VancheeZze: Renamed

packages/bodiless-organisms/src/components/FilterByGroup/FilterByGroupClean.tsx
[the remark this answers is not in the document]
wodenx: right - so i'm asking whether we need to provide these default texts as props.

examples/test-site/src/components/Page/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { flow } from 'lodash';
+import { Page } from '@bodiless/gatsby-theme-bodiless';
+import { withPageDimensionsContext } from '@bodiless/components';
+
+const breakpoints = {
+  sm: 640, md: 768, lg: 1024, xl: 1280,
```
hvanyo, line 20: @VancheeZze  This seems weird to define there here as they really come from and controlled by Tailwind config.   Is it possible to use something like withTailwindClasses() and bring these values in from the Tailwind.config instead of defining them here.   Eric used it data/pages/flowcontainer page. 
Otherwise a site developer has to change them both in tailwindconfig and in this code which is not ideal.

examples/test-site/src/components/FlowContainer/token.tsx
```diff
@@ -29,7 +29,7 @@ const asFlowContainerFullWidth = withDesign({
 });
 
 const asFlowContainerFilterable = withDesign({
-  ComponentWrapper: asFilterableByGroup,
+  ComponentWrapper: asFilterableByGroup(),
```
wodenx, line 32: I think this is not generic enough - what if the item you're wrapping already has a '_tags" child?  So, we should either make this token specific to product touts (or touts in general), or we should give this function a nodeKey parameter.

VancheeZze: `asFilterableByGroup` is already accepts `nodeKey` which is `_tags` by default, but we could've done asFilterableByGroup('new_node_key')

packages/bodiless-components/src/Accordion/AccordionBody.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { expanded } = useAccordionContext();
+
+  return (
+    <Wrapper className={expanded ? 'block' : 'hidden'}>
+      { children }
+    </Wrapper>
+  );
+};
+
+const asExpandedOnDesktop = ifViewportIs(['sm', 'md'])(
```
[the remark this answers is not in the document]
VancheeZze: Fixed

packages/bodiless-components/src/Accordion/AccordionBody.tsx
```diff
@@ -0,0 +1,65 @@
...
+const AccordionBodyComponentsStart:AccordionBodyComponents = {
+  Wrapper: Div,
+};
+
+const AccordionBodyBase: FC<AccordionBodyProps> = ({ components, children }) => {
+  const { Wrapper } = components;
+  const { expanded } = useAccordionContext();
+
+  return (
+    <Wrapper className={expanded ? 'block' : 'hidden'}>
```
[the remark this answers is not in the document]
VancheeZze: Fixed

packages/bodiless-components/src/Accordion/AccordionTitle.tsx
```diff
@@ -46,8 +49,15 @@ const AccordionTitleBase: FC<AccordionTitleProps> = ({ components, children }) =
...
 
+const asNoIconOnDesktop = ifViewportIs(['sm', 'md'])(
+  withDesign({
+    Icon: addClasses('lg:hidden'),
+  }),
+);
+
 const AccordionTitleClean = flow(
   designable(AccordionTitleComponentsStart),
+  asNoIconOnDesktop,
```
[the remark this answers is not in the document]
VancheeZze: Fixed

examples/test-site/src/components/FilterByGroup/token.tsx
```diff
@@ -41,12 +42,16 @@ const withTagListStyles = withDesign({
 
 const withCategoryList = withDesign({
   Title: addClasses('font-bold'),
+  Wrapper: flow(
```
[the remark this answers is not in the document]
VancheeZze: Fixed. The initial list items were defined as `asComponent('li')` which is not stylable by default. Replaced it with `Ul` and `Li` elements from the `@bodiless/fclasses`

examples/test-site/src/components/Elements.token.ts
```diff
@@ -36,6 +36,7 @@ const asXMargin = addClasses('mx-2');
 const asYMargin = addClasses('my-2');
 const asNegXMargin = addClasses('-mx-1');
 const asNegYMargin = addClasses('-my-1');
+const asWithPadding = addClasses('p-5');
```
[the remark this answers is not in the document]
VancheeZze: Fixed

examples/test-site/tailwind.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   */
 
   // prefix: '',
-  // important: false,
+  important: '#___gatsby', // https://tailwindcss.com/docs/configuration/#important
```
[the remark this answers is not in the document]
VancheeZze: So some our 3rd party CSS classes were impossible to overwrite because of tailwind css loading first and then 3rd party css was added below it making tw selectors have lower priority. 

> This configuration will prefix all of your utilities with the given selector, effectively increasing their specificity without actually making them !important.

In this case, It was needed to be able to hide the accordion icon with `md:hidden` class on the desktop. Otherwise, `material-icon` classes were taking precedence.

packages/bodiless-components/src/List/index.tsx
```diff
@@ -21,7 +21,7 @@ import React, {
 import { flow } from 'lodash';
 import { observer } from 'mobx-react-lite';
 import { withNode, withoutProps } from '@bodiless/core';
-import { designable, asComponent } from '@bodiless/fclasses';
+import { designable, Ul, Li } from '@bodiless/fclasses';
```
wodenx, line 24: Sorry I think I misled you here - we probably don't want to use these in @bodiless/components - so my question about why the wrapper wasn't already stylable just needed an answer, not a change...  Can you revert this please?

VancheeZze: Fixed

Issue #226, opened 2020-03-23
Migrate page and resource redirections


schristodoulou:
## User Story

As a site flattener tool operator
I want to migrate page and resource redirections
So that I can preserve existing site behavior

### Acceptance Criteria

AC1
when migration tool hits pageA that is redirected to pageB
then the redirection is written to a yaml file

AC2
when migration tool hits pageA that is redirected to pageB
then the redirection is logged to the terminal

AC3
when migration tool hits pageA that is redirected to pageB
then pageA is not created
then pageB is created

AC4
when migration tool hits pageA that is redirected to pageB hat is redirected to pageC, etc (pageA -> pageB -> pageC -> ... -> pageN)
then pageA, pageB ... pageN-1 are not created
then pageN is created if the server responded with Success status

AC5
when migration tool hits resourceA (image, style, script, etc) that is redirected to resourceB by the server
then the redirection is written to .platform/routes.yaml according to psh doc

AC6
when migration tool hits resourceA (image, style, script, etc) that is redirected to resourceB by the server
then the redirection is logged to the terminal

AC7
when migration tool hits resourceA (image, style, script, etc) that is redirected to resourceB by the server
then resourceeA is not created
then resourceB is created

AC8
when migration tool hits resourceA that is redirected to resourceB that is redirected to resourceN, etc (resourceA -> resourceB -> resourceC -> ... -> resourceN)
then resourceA, resourceB ... resourceN-1 are not created
then resourceN is created if the server responded with Success status

AC9
unit tests created

AC10
documentation is updated how to include results from AC1 to psh routes

hvanyo:
@dewen  @beliayeu  @schristodoulou  @TatsianaMiraniuk 
The downloading of resources is being done in two ways.  
This solution above is being implemented on the first method.
If the resources are attempted to be downloaded and found missing in the second method, they will not be redirected with this solution.  We will put these gaps tickets to determine amount of cases and if we should add this solution to second method.    i.e. srcset in pictures
@dewen  @beliayeu  @TatsianaMiraniuk  -- Tanya will probably need help to determine if we are following into the second case and missed  OR its the first case and not working.

PR #309, opened 2020-05-01
feat(Migration tool): Migrate page and resource redirections.

dewen:
## Changes

Add feature for migrate page and resource redirections

### a. Create redirect rules from migrating site.
- Clone repo and checkout the branch
- npm run setup
- Go to example starter and configure migration-settings.json
```
{
  "url": "http://aesq.ucxp.com/",
  "crawler": {
    "maxDepth": 2,
    "maxConcurrency": 1,
    "ignoreRobotsTxt": true
  },
  "transformers": [],
  "exports": {
    "redirects": {
      "path": "/path/to/redirect_export.yaml",
      "format": "yaml"
    }
  }
}
```
- npm run migrate
- check the exported redirect rules inside /path/to/redirect_export.yaml

### b. Test redirect on platform.sh (psh)
- Create/clone starter project that can be deployed on psh.
- Copy over the migrated content
- Update router.yaml with redirect rules.
- Deploy on psh
- Test page redirects.

## Related Issues

Fixes: #226


packages/bodiless-migration-tool/src/site-flattener.ts
```diff
@@ -166,7 +181,45 @@ export class SiteFlattener {
       );
       await downloader.downloadFiles([fileUrl]);
     });
+    const exports = this.params.exports || null;
+    if (!!exports.redirects) {
+      const redirectKeys: Array<string> = [];
+      scraper.on('responseReceived', async response => {
+        if ([301, 302, 307, 308].indexOf(response.status()) !== -1) {
```
beliayeu, line 188: I believe in order to make it more maintainable and testable it would be great to move this logic to a separate function or even file

packages/bodiless-migration-tool/src/site-flattener.ts
```diff
@@ -166,7 +181,45 @@ export class SiteFlattener {
...
+        }
+      });
+
+    }
     await scraper.Crawl();
+
+    // Export redirect rules.
+    if (redirects.length && exports.redirects.path) {
+      const format = exports.redirects.format || 'yaml';
+      switch(format) {
```
beliayeu, line 209: it  would be great to add some validation, probably if invalid format is specified, then we can fall back to yaml

packages/bodiless-migration-tool/src/site-flattener.ts
```diff
@@ -166,7 +181,45 @@ export class SiteFlattener {
...
+    if (!!exports.redirects) {
+      const redirectKeys: Array<string> = [];
+      scraper.on('responseReceived', async response => {
+        if ([301, 302, 307, 308].indexOf(response.status()) !== -1) {
+          const headers = response.headers();
+          const from = this.getRedirectPath(response.url());
+          if (redirectKeys.indexOf(from) === -1) {
+            redirectKeys.push(from);
+            const redirectRule: RedirectRule = {};
+            redirectRule[from] = {
```
beliayeu, line 194: do we handle duplicates?

dewen: These will be repeated redirects, so this logic is to make redirect from unique:
`if (redirectKeys.indexOf(from) === -1) {`

packages/bodiless-migration-tool/src/site-flattener.ts
```diff
@@ -166,6 +173,17 @@ export class SiteFlattener {
...
       await downloader.downloadFiles([fileUrl]);
     });
+    scraper.on('responseReceived', async response => {
+      if (!isEmpty(exports) && !isEmpty(exports.redirects)) {
+        responseProcessor.processRedirect(response);
+      }
+    });
+    scraper.on('requestFinished', async () => {
+      if (!isEmpty(exports) && !isEmpty(exports.redirects)) {
+        responseProcessor.exportRedirects(exports.redirects);
```
beliayeu, line 183: I believe we are doing redirects flushing (writing data to fs) too often. it may lead to performance degradation  and not sure we want to do it so often. As an option we can do it after a page is scraped or just once after all pages are scraped

packages/bodiless-migration-tool/src/response-processor.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  public processRedirect(response: Response): void {
+    if ([301, 302, 307, 308].indexOf(response.status()) !== -1) {
+      const headers = response.headers();
+      const from = ResponseProcessor.getRedirectPath(response.url()).replace(/\/$/g, '');
+      const destUrl = headers.location;
+      const destination = isUrlExternal(response.request().url(), destUrl)
+        ? headers.location
+        : ResponseProcessor.getRedirectPath(headers.location);
```
beliayeu, line 67: have you considered using requestredirectchain method https://github.com/puppeteer/puppeteer/blob/v3.0.2/docs/api.md#requestredirectchain

packages/bodiless-migration-tool/src/site-flattener.ts
dewen: I moved this to response-processor.ts

packages/bodiless-migration-tool/src/response-processor.ts
dewen: I can give a try. 

But if it's a redirect to external resource, we might not be able to catch the redirect chain.

packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -109,14 +110,32 @@ export default class Downloader {
...
+          return false;
+        }
+        return true;
+      }
       // @ts-ignore retryRequest does not have type definition for the function
       // that produces request.Request.
-      const req: request.Request = retryRequest({ uri: resource });
+      const req: request.Request = retryRequest({
+        uri: resource,
+        followRedirect,
```
beliayeu, line 128: per my understanding default behavior is to follow HTTP 3xx responses as redirects. are there reasons why do we want to override the default behavior?

packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -109,14 +110,32 @@ export default class Downloader {
...
+      });
+
       req
         .on('response', res => {
           if (res.statusCode >= 400) {
             return reject(new Error(`Resource ${resource} is not available for download.`));
           }
+          if ([301, 302, 307, 308].indexOf(res.statusCode) !== -1) {
```
beliayeu, line 136: as we follow redirections (followRedirect = true), under what circumstances do we get a case when status code is 3xx ?

packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -110,32 +109,21 @@ export default class Downloader {
...
           }
-          if ([301, 302, 307, 308].indexOf(res.statusCode) !== -1) {
-            return reject(new Error(`Resource ${resource} is redirected. No download.`));
+          if (res.request.href !== resource) {
+            targetPath = this.getTargetPath(res.request.href);
+          }
+          if (targetPath === undefined) {
+            return Promise.reject(new Error(`target path for ${res.request.href} is undefined`));
           }
+          debug(`Resource ${resource} redirect to ${res.request.href}.`);
```
beliayeu, line 126: I believe we should either remove this debug or change this message since we may get resources here without redirections

packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -100,7 +99,7 @@ export default class Downloader {
   }
 
   private downloadFile(resource: string) {
-    const targetPath = this.getTargetPath(resource);
+    let targetPath = this.getTargetPath(resource);
```
beliayeu, line 102: I think we can remove the logic from here. I think it is sufficient to have inside .on('response'

packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -117,6 +117,16 @@ export default class Downloader {
           if (res.statusCode >= 400) {
             return reject(new Error(`Resource ${resource} is not available for download.`));
           }
+          if (res.request.href !== resource) {
+            const redirectTargetPath = this.getTargetPath(res.request.href);
```
beliayeu, line 121: @dewen, I think the above logic related to target path can be removed - https://github.com/johnsonandjohnson/Bodiless-JS/blob/bbba64911e193b63732d7369e367d7ba606b1310/packages/bodiless-migration-tool/src/downloader.ts#L102-L110

PR #310, opened 2020-05-01
test(Components, test-site): better selectors for List

awm086:


packages/bodiless-components/src/List/index.tsx
```diff
@@ -74,6 +76,18 @@ const BasicList: FC<Props> = ({ components, unwrap, ...rest }) => {
   );
 };
 
+const asTestableList = withDesign({
+  Wrapper: addProps({ 'data-list-element': 'list' }),
+  Title: addProps({ 'data-list-title': 'list-title' }),
+  Item: addProps({ 'data-list-item': 'list-item' }),
```
VancheeZze, line 82: I'd keet `data-attr` all with the same name and different values:

```js
  Wrapper: addProps({ 'data-list-element': 'wrapper' }),
  Title: addProps({ 'data-list-element': 'title' }),
  Item: addProps({ 'data-list-element': 'item' }),
```

PR #315, opened 2020-05-05
fix(migration-tool): 'TypeError: tag[primaryAttributeKey].toLowerCase is not a function' error during a build

beliayeu:
## Changes
* updated the logic that converts html element into Helmet element. When source html element text is empty then the attribute of the Helmet element is not created.

## Notes
* the error is thrown by Helmet. The reason is we may end up with empty Helmet attribute after the migration. The empty Helmet attribute is caused by html elements with empty text. Hereby, the logic that prepares Helmet element is updated, particularly, the attribute is not created when element text is empty.


packages/bodiless-migration-tool/src/html-parser.ts
```diff
@@ -312,6 +312,10 @@ export default class HtmlParser implements HtmlParserInterface {
     if (tag === 'script' && this.$(element).attr('src')) {
       skip = true;
     }
+    // skip if html element text is empty
+    if (!this.$(element).html()) {
```
dewen, line 316: This check might affect other element behavior, i.e. <style>, new 'cssText' attribute will not be created if it's empty.

Is this expected behavior?

beliayeu: yep, that is expected. style and cssText has the same problem.

PR #352, opened 2020-05-26
feat(core, components): Improve/Document context menu form API and Editable

### LINKED ISSUE #331: Improve and Document ContextMenuForm API
## User Story
As a developer
I want a well documented and easy to use API for creating forms
So I can be more productive

## Acceptance Criteria
AC1. There is a guide for building a custom component form and menu button, including both high-level API  (`withEditButton`) and lower level (`ContextMenuForm` and `withMenuOptions`).
AC2. The `ContextMenuForm` component is exposed for direct use,  with all options available as props.
AC3. Tests are written for `ContextMenuForm`
AC4. current uses of `contextMenuForm` are refactored to make use of new API and Informed hooks.

### LINKED ISSUE #335: Improve Editable and asEditable
## User Story
As a developer
I want improvements to Editable
So I can apply basic edit functionality more easily and flexibly.

## Acceptance Criteria
AC1.
The `asEditable` HOC can take an object for its first parameter
Exactly as described in #334 

AC2.
A component wrapped. in `asEditable` allows the following props
To be overridden when the component is used:
`nodeKey`, `nodeCollection` and `placeholder`

AC3.
A component wrapped in asEditable
does not render an unnecessary span in preview or static mode.

AC4.
The default content for a component wrapped in `asEditable`
will be taken from its children prop
eg
```
const Span = asEditable('foo')('span');
<Span>Hello</Span>
```
Will render "Hello" when there is no data at nodeKey 'foo'.

wodenx:
## Changes
- Add asBodilessComponent API
- Refactor asEditable to allow nodekey and render children as default content.
- Put context menu form ui in a context
- Refactor ContextMenuForm so that submit handler, initial values, etc can
  be provided as props.
- Change the return value of context menu handlers to a render function (not
  a component).

## Related Issues
Closes #331 
Closes #335


examples/test-site/src/data/pages/forms/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { asEditable } from '@bodiless/components';
+import ReactMarkdown, { ReactMarkdownProps as Props } from 'react-markdown';
+
+import Layout from '../../../components/Layout';
+import MarkdownField from './InformedMarkdown';
+
+type Data = Pick<Props, 'source'>;
+
+const asBodilessMarkdown = asBodilessComponent<Props, Data>({
+  icon: 'edit',
```
beliayeu, line 27: consider having a dedicated prop that will store all context menu settings

examples/test-site/src/data/pages/forms/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  name: 'edit',
+  renderForm: () => {
+    // const { ComponentFormLabel, ComponentFormTitle, ComponentFormTextArea } = useFormUI();
+    const { ComponentFormLabel, ComponentFormTitle } = useFormUI();
+    return (
+      <>
+        <ComponentFormTitle>Markdown</ComponentFormTitle>
+        <ComponentFormLabel>Content</ComponentFormLabel>
+        {/* <ComponentFormTextArea field="source" /> */}
+        <MarkdownField field="source" />
```
beliayeu, line 37: just curious if there is a way for site builders to manipulate data before saving it to storage and after reading it from storage. for instance, it may be a use cases in which they need to implement input sanitization

also, is there a way to add some custom validation rules?

packages/bodiless-components/src/Editable.tsx
```diff
@@ -84,14 +83,43 @@ const withPlaceholder = <P extends object> (placeholder?: string) => (Component:
...
+      } = props;
+      const editableProps = {
+        children,
+        placeholder: placeholderProp || placeholder,
+        nodeKey: nodeKeyProp || nodeKey,
+        nodeCollection: nodeCollectionProp || nodeCollection,
+      };
+      return (
+        <Component {...rest as P}>
+          <Editable {...editableProps} />
```
beliayeu, line 116: am I right that we do not allow sending props different to editableProps to Editable? so as a site builder when I have a component composed using asEditable and I want to add a class to Editable, then I will have to recompose it

examples/test-site/src/data/pages/forms/index.tsx
wodenx: @beliayeu Not sure what you mean here.  Let's discuss.

packages/bodiless-components/src/Editable.tsx
wodenx: That is correct, but there should be no need to add a class to editable - in fact, in the ideal world, editable would just be a fragment when not in edit mode (it's currently a span, for some reason I can't remember).  But any styling, etc should be done on the enclosing element.

examples/test-site/src/data/pages/forms/index.tsx
wodenx: >  site builders to manipulate data before saving it to storage and after reading it from storage.
Currently they can do either one by using a lower-level API and implementing their own `submitValues()` method and/or processing the data from the node before setting `initialValues`.  I was planning on adding an example of using the lower level API.  At some point, we can consider enhancing the high-level `asBodiless...` API to make this easier, but not sure how critical it is at this point.


> is there a way to add some custom validation rules?
> 
> will it possible to do it using informed/asField?

You mean for custom fields?  I think so, yes - see [the informed docs](https://joepuzzo.github.io/informed/?path=/story/custominputs--creating-custom-inputs).

I think we already have examples of adding validation for existing informed fields in [NewPageProvider](https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/gatsby-theme-bodiless/src/dist/NewPageProvider.tsx#L99)

wodenx: @beliayeu can this convo be resolved?

packages/bodiless-components/src/Editable.tsx
wodenx: @beliayeu can this convo be resolved?

packages/bodiless-components/src/Editable.tsx
```diff
@@ -36,7 +35,7 @@ type Data = {
 const Text = observer((props: Props) => {
   const { placeholder } = props;
   const { node } = useNode<Data>();
-  const text = node.data.text || placeholder || '';
+  const text = node.data.text || props.children || placeholder || '';
```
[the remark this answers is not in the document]
wodenx: @e2thex yes, i think so - and i believe that is already enabled by the work @beliayeu  did on image placeholders.  The only requirement is that you specify the schema (as a generic default value) in `withNodeDataHandlers` so that we know which props to pick to add to the default data.

packages/bodiless-components/src/Editable.tsx
```diff
@@ -36,7 +39,7 @@ type Data = {
 const Text = observer((props: Props) => {
   const { placeholder } = props;
   const { node } = useNode<Data>();
-  const text = node.data.text || placeholder || '';
+  const text = (node.data.text !== undefined ? node.data.text : props.children) || placeholder || '';
```
beliayeu, line 42: consider doing it as
```
const text = node.data.text || props.children || placeholder || '';
```

wodenx:
This was already merged.

PR #369, opened 2020-06-08
feat(documentation): allow to override doc site resources locally

beliayeu:
## Changes
* allowed to override doc site resources locally
* moved toc configuration from bodiless.docs.toc.js into bodiless.docs.settings.js


dewen:
Please add the fixing issue #.

PR #373, opened 2020-06-09
feat(components): remove the # from the link URL input field

### LINKED ISSUE #374: Remove the # from the link URL input field
## User Story

As a Content Editor, I want to add a link to my content so that my Users can navigate through the site.

As a Content Editor, When I use the Link component to create my link, I want a clear indication of where to add the link and how to add it, so there is no confusion with the Hashtag

### Acceptance Criteria

AC1:

The link component no longer defaults with the "#"

AC2:

The link component defaults with a placeholder that can be set, "/link"

AC3:

The component includes information text below the link field

"Use relative URLs for internal links. Preface the link with '/' to be relative to the root, otherwise the link is relative to the page. 
Use a fully formed URL for external links, e.g., https://www.xyz.com."

AC4:

If their is no value added by the editor the field will be considered empty by the system and render a #

If the user enters a value in place of the /link then the page will render the corresponding value. 

QA notes:
test cases and test results
 

### Implementation Notes

Tech Notes:

Test the behavior of the "remove link" functionality

awm086:
## Changes
- Remove the # from the link URL input field.

## Test Instructions
- Add a link to editable text.
- Open the link form.
- Note the link initial value

## Related Issues
Closes #374


packages/bodiless-components/src/Link.tsx
```diff
@@ -80,7 +86,7 @@ export const editButtonOptions: EditButtonOptions<Props, Data> = {
 };
 
 const emptyValue = {
-  href: '#',
+  href: '/link',
```
awm086, line 89: @beliayeu I am unable to make this meet ac 4. The empty value is what get saved and what displays in the form.  I tried adding a placeholder, but that would only work if the emptyValue is empty string and since we default to `#` or `/link` it's hard to seperate what appear in the form field and what get saved. I tried using formApi with a bit of logic to force the value prop on the field : 
```
formApi.setValue('href', calculatedValue)
```
but that did not work.

beliayeu: @awm086, am I right that we want to implement the following case:
when link value is empty or does not exist in bodiless store
then value of component form text field should be empty
then link (href attribute of 'a' html tag) should have '#'  ?

if I understand it correctly, then we can implement it the following way - https://github.com/johnsonandjohnson/Bodiless-JS/compare/master...beliayeu:chore/link-direction?expand=1

it would be great to implement this in reusable way since there can be other use cases in which we want to process bodiless data before displaying it to end-user (for instance, input sanitization)

also, it would be great to include @wodenx and @e2thex to the conversation for architectural review.

packages/bodiless-components/src/Link.tsx
```diff
@@ -80,8 +87,21 @@ export const editButtonOptions: EditButtonOptions<Props, Data> = {
 };
 
 const emptyValue = {
-  href: '#',
+  href: '',
+};
+
+const withPropTransformer = (transformer: Function) => (prop: string) => (
```
wodenx, line 93: After discussing with @beliayeu , I don't think we actually need this abstraction here - I think just
```
const wiithHrefTransformer = Component => ({ href, ...rest }) => <Component href={href !== '' ? href : '#'} {...rest} />
```
is sufficient (with proper types added, of course)

wodenx: Also - with a display name.

PR #387, opened 2020-06-17
[DO NOT MERGE] bug(richtext, core): Prevent unnecessary re-activation of context for slate editor.

### LINKED ISSUE #379: Unnecessary re-resolve of slate editor props
## STR
1. Load test-site homepage
2. Put into edit mode
3. Click on CTA link in first Tout
4. Click link button
5. Observe browser console

## ER
No warnings

## AR
```
react_devtools_backend.js:6 A Slate <Editor> component is re-resolving the `plugins`, `schema`, `commands`, `queries` or `placeholder` prop on each update, which leads to poor performance. This is often due to passing in a new references for these props with each render by declaring them inline in your render function. Do not do this! Declare them outside your render function, or memoize them instead. 
```

wodenx:
## Changes
- Memoize props provided to slate editor by RichText.
- Prevent reactivation of a context when it is already innermost.
- Modify RichText so that it does not refresh context on every change.
- Update LocalContextMenu so it only re-renders when it is actually on-screen.

## Test Instructions
- Visit a page with many slate editors
- Click on one slate editor many times and/or type into one
- Observe whether warnings appear in console about re-resolving slate editor props, and whetehr performance of slate editor is acceptable.

## Related Issues
Fixes #379

### DIFF
--- a/packages/bodiless-core/src/components/LocalContextMenu.tsx
+++ b/packages/bodiless-core/src/components/LocalContextMenu.tsx
@@ -61,6 +61,9 @@ const LocalContextMenu: FC = ({ children }) => {
   // let the context know it has a localMenu
   context.hasLocalMenu = true;
   const { isInnermostLocalMenu, areLocalTooltipsDisabled } = context;
+  // TODO: Only render tooltip when needed. Currently this causes focus issues with editables.
+  // (The editable gets the focus, then the tooltip re-renders and creates a new editable
+  // which is not focused.  Might be worth investigating this at some point.)
   // if (!isInnermostLocalMenu || areLocalTooltipsDisabled) {
   //   return <>{children}</>;
   // }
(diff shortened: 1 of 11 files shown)


packages/bodiless-core/src/components/LocalContextMenu.tsx
```diff
@@ -20,28 +20,54 @@ import { useEditContext } from '../hooks';
...
-    if (element.getBoundingClientRect().left <= 0) {
-      element.style.visibility = 'hidden';
-    } else {
-      element.style.visibility = 'visible';
-    }
-  };
+  const { isInnermostLocalMenu, areLocalTooltipsDisabled } = context;
+  // if (!isInnermostLocalMenu || areLocalTooltipsDisabled) {
+  //   return <>{children}</>;
+  // }
```
VancheeZze, line 66: Will we consider going back to not render a tooltip at all in a future or we can just remove these comments and the one below?

wodenx: Addressed this in https://github.com/johnsonandjohnson/Bodiless-JS/pull/387/commits/a296e08cf073e7cf55c53daa82d0fd50c3564145

wodenx:
Closing as this was included in last round of dependency updates.

PR #409, opened 2020-06-29
Feat: implement interface for managing meta data

dewen:
## Changes

- Enhanced context menu form.
- Compound form API.
- Updated withMeta to support meta data editing and rendering.
- Test site layout update

## Test Instructions

- Context menu form regression
- Meta SEO tag rendering

## Related Issues

Fixes: #450


examples/test-site/src/components/Layout/index.jsx
```diff
@@ -72,7 +135,11 @@ const Layout = ({ children }) => (
         <ExampleHelmet />
         <ExampleGTMHelmetEvent />
         <Header siteLogo={data.site.siteMetadata.logo} />
-
+        <Seo>
```
wodenx, line 138: i would rather see the form snippets applied in exactly the same place that the withMeta are applied -- in fact, i'd consider modifying withMeta so that it adds the form snippets when in edit mode. 

I think you might be able to refactor withMeta to use the new sidecar node api, something like
```
const withMeta$ = (name: string) => (HelmetComponent: CT) => (props: any) => {
  const { children, content, ...rest } = props;
  if (content) {
    return (
      <HelmetComponent {...rest}>
        {children}
        <meta name={name} content={content} />
      </HelmetComponent>
    );
  }
  return <HelmetComponent {...rest} />;
};

// Note, in the new nodeKey api, nodeKeys can be an object which includes both nodeKey and nodeCollection
const withMeta = (name, nodeKeys, defaultData = { content: undefined }) => withSidecarNodes(
    withNodeKey(nodeKey),, 
    withNode,
     withNodeDataHandlers(defaultData),
     ifEditable(
       withEditFormSnippet(() => <SeoSnippet {...defaultData} />),'
    ),
    withoutProps('setComponentData'),
    withData,
    withMeta$(name),
);
```

wodenx: And then add withCompoundForm to asBodilessHelmet.

The idea of all of this is to keep the edit UI as close as possible to the components that use the data.

examples/test-site/src/components/Layout/index.jsx
```diff
@@ -45,50 +47,107 @@ const useGetMenuOptions = () => () => [
...
+  return WithSeoFormHeader;
+};
+
 const Seo = flowRight(
-  withCompoundForm({ useGetMenuOptions, name: 'Seo', peer: true }),
+  withCompoundForm({
+    useGetMenuOptions, name: 'Seo', peer: true, id: 'seo',
+  }),
+  withSeoFormHeader(),
 )(Div);
```
wodenx, line 80: We definitely don't want to apply this to a Div.

examples/test-site/src/components/Layout/index.jsx
```diff
@@ -45,50 +47,107 @@ const useGetMenuOptions = () => () => [
...
   withoutProps('setComponentData'),
   withData,
 );
 
-const Comp = () => (
-  <></>
-);
+const Comp = () => <></>;
 
 const MetaTitle = flowRight(
```
wodenx, line 118: You still need to combine the form definition with the withMeta HOC which renders the data.  I think you have 2 choices here.
1. use the sidecar node pattern from my previous comment, and apply everything to a single Helmet component. 
2. use a separate Helmet component for each metatag.
The former has the advantage that you could also apply the form header to the helmet component, and not need a separate `<Seo>` (probably bettern named `<SeoForm>` or `<MetaForm>`) tag.  And this is more consonant with our current patterns of exporting HOC's.
Either way, I think you need to refactor withMeta so as to remove [the direct access to the child node](https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-components/src/Meta/Meta.tsx#L30), bc you'll want it ultimately to be something like:
```
const withMeta = (nodeKey, optioons) => 
flowRight(
  withNodeKey(nodeKey),
  withNode,
  withNodeDataHandlers,
  ifEditable(
    withMetaSnippet(options),
  ),
  withMetaChild(options) // This is where we are actually doing the work of adding the metatag ito the helmet
);
```
where `withMetaChild` (or whatever) is [this logic](https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-components/src/Meta/Meta.tsx#L30).

packages/bodiless-core/src/withEditFormSnippet.tsx
```diff
@@ -9,11 +9,22 @@ import type { FormBodyRenderer as Renderer } from './withEditButton';
...
+  fromJSON?: (jsonValues: any) => D,
+};
+
+const withEditFormSnippet = <P extends object, D extends object>(options: Options<P, D>) => (
   (Component: CT<P>) => {
     const id = v1();
+    const { render, fromJSON, toJSON } = options;
     const WithEditFormSnippet = (props: P & EditButtonProps<D>) => {
       const { unwrap } = props;
+      const { submitValues, initialValues } = useEditFormProps(props);
```
wodenx, line 24: I would move this logic to useEditFormProps so it is available to normal forms as well.

examples/test-site/src/components/Layout/index.jsx
```diff
@@ -45,50 +47,107 @@ const useGetMenuOptions = () => () => [
...
+
+const SeoHeaderSnippet = {
+  id: v1(),
+  render: () => {
+    const { ComponentFormTitle, ComponentFormDescription } = useFormUI();
+    return (
+      <>
+        <ComponentFormTitle>SEO Data Management</ComponentFormTitle>
+        <ComponentFormDescription>
+          {`Enter the page level data used for SEO.
```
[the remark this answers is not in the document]
dewen: fixed.

examples/test-site/src/components/Layout/index.jsx
```diff
@@ -45,50 +47,107 @@ const useGetMenuOptions = () => () => [
...
-  withEditFormSnippet(() => <SeoSnippet {...defaultData} />),
+  withMetaSnippet(defaultData),
   withoutProps('setComponentData'),
   withData,
 );
 
-const Comp = () => (
-  <></>
-);
+const Comp = () => <></>;
```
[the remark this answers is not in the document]
dewen: Applied to Meta component from @bodiless/fclasses

Is it ok?

packages/bodiless-core/src/withEditFormSnippet.tsx
```diff
@@ -11,20 +11,16 @@ import type { Snippet } from './withCompoundForm';
 
 type Options<P, D> = {
   render: Renderer<P, D>,
-  toJSON?: (values: D) => any,
-  fromJSON?: (jsonValues: any) => D,
+  submitValueHandler: (values: D) => any,
+  initialValueHandler: (jsonValues: any) => D,
```
beliayeu, line 15: @dewen, I think we should rename jsonValues

examples/test-site/src/data/pages/compound-form/index.tsx
```diff
@@ -65,7 +65,7 @@ const asTestFormSnippet = (nodeKey: WithNodeKeyProps, defaultData: any) => flowR
   withNodeKey(nodeKey),
   withNode,
   withNodeDataHandlers(defaultData),
-  withEditFormSnippet(() => <TestSnippet {...defaultData} />),
+  withEditFormSnippet({ render: () => <TestSnippet {...defaultData} /> }),
```
wodenx, line 68: Maybe consider an overload here, since in many cases you won't need custom submit handlers.

packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -12,15 +12,58 @@
...
+    {children}
+    <meta name={name} content={content} />
+  </HelmetComponent>
+);
+
+/**
+ * Meta data options. Set the data name, content, form field and label.
+ * @param options Options
+ */
+const withMeta = (options: Options) => (nodeKey?: WithNodeKeyProps) => withSidecarNodes(
```
wodenx, line 56: Rather than having `content` in the options, I think it is more consistent with other bodiless api's to have it as a 2nd param to the HOC factory this returns, so
```
const withMeta = (options: Options) => (nodeKey?: WithNodeKeyProps, defaultContent?: string) => ...
```
see [`AsBodiless`](https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-core/src/asBodilessComponent.tsx#L36)

packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -12,15 +12,58 @@
...
+  withNodeKey(nodeKey),
+  withNode,
+  withNodeDataHandlers(pick(options, 'content')),
+  ifEditable(withMetaSnippet({ ...options, attribute: 'content' })),
```
wodenx, line 60: I think you should allow overriding the attribute in the options.   Also, in the line above, you should use the attribute to define the data, so
```
withNodeDataHandlers({ [options.attribute || 'content']: defaultContent }),
ifEditable(withMetaSnippet({ attribute: 'content', ...options }),
```

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return (
+      <Div key={name}>
+        <ComponentFormLabel>{label}</ComponentFormLabel>
+        <Field field={name} />
+      </Div>
+    );
+  },
+  submitValueHandler: (values: any) => {
+    const { name, attribute } = data;
+    if (next) {
```
wodenx, line 41: The idea of `next` is to *chain* the handlers - so you should always do this manipulation and then pass the transformed data on to next, if it is specified.  Also, you probably need both  `nextSubmitHandler` and a`nextInitialValuesHandler`as custom transformations will have to happen in both directions.

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  const WithFormHeader = (props: any) => {
+    useRegisterSnippet(metaHeaderSnippet);
+    return <Component {...props} />;
+  };
+  return WithFormHeader;
+};
+
+const withMetaForm = (useGetMenuOptions: any, metaFormHeader: HeaderProps) => flowRight(
```
wodenx, line 79: I think we should decompose this.  I also think we should provide defaults for withMetaFormHeader so they don't have to be specified at site level.

packages/bodiless-core-ui/src/GlobalContextMenu.tsx
```diff
@@ -25,7 +25,7 @@ import {
   ComponentFormTitle, ComponentFormCloseButton, ComponentFormLabel, ComponentFormText,
   ComponentFormButton, Icon, Div, Hr, ToolbarButton, ComponentFormUnwrapButton,
   ComponentFormError, ComponentFormSubmitButton, ComponentFormList, ComponentFormListItem,
-  ComponentFormDescription, ComponentFormTextArea,
+  ComponentFormDescription, ComponentFormTextArea, ComponentFormWarning,
```
wodenx, line 28: Is this intentional?  Where do we use this?

dewen: This is merged from master, from resolved conflict

examples/test-site/src/components/Layout/index.jsx
```diff
@@ -1,5 +1,5 @@
 /**
```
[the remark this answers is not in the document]
dewen: @hvanyo  Currently, ExampleHelmet is the component handling meta data related logic, I can move it to meta.tsx if this is what you suggest.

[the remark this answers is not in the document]
hvanyo: My suggestion here was to create Layout/meta.tsx and move lines 33 - 78 to the file and then import ExampleHelmet.
Just trying to keep the meta components separate from layout.

examples/test-site/src/components/Layout/meta.tsx
```diff
@@ -0,0 +1,72 @@
...
+const seoFormHeader = {
+  title: 'SEO Data Management',
+  description: `Enter the page level data used for SEO. 
+  This is metadata needed for SEO that will go in the page header.`,
+};
+
+const SiteHelmet = flowRight(
+  withMetaForm(useGetMenuOptions, seoFormHeader),
+  asBodilessHelmet('meta'),
+  withMetaTitle('page-title', 'Rec 30-65 char'),
```
hvanyo, line 64: @dewen  for them to understand page vs site, I would rename these.

withMetaPageTitle
withMetaPageDescription
withMetaPageType (correct)

packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -36,28 +32,36 @@ type Props = BaseProps & Data;
...
 
 const withMeta$ = (name: string) => (
   HelmetComponent: CT<BaseProps>,
-) => ({ children, content, ...rest }: Props) => (
-  <HelmetComponent {...rest}>
-    {children}
-    <meta name={name} content={content} />
-  </HelmetComponent>
-);
+) => ({ children, content, ...rest }: Props) => ((name === 'title')
```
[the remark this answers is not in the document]
wodenx: Actually -now that I look at the code, I think that's too much for the amount of reuse you're getting.
I'd just  split this into withMeta$ and withTitle$, like:
```
const withTitle$ = () => (
  HelmetComponent: CT<BaseProps>,
) => ({ children, content, ...rest }: Props) => (
  <HelmetComponent {...rest}>
    {children}
    <title>{content || ''}</title>
  </HelmetComponent>
);

const withMeta$ = (name: string) => (
  HelmetComponent: CT<BaseProps>,
) => ({ children, content, ...rest }: Props) => (
 : (
    <HelmetComponent {...rest}>
      {children}
      <meta name={name} content={content} />
    </HelmetComponent>
  ));
```
and then something like:
```
const withHeadElement = (renderHoc) => (options: Options) => (
  nodeKey?: WithNodeKeyProps, defaultContent?: string,
) => withSidecarNodes(
  withNodeKey(nodeKey),
  withNode,
  withNodeDataHandlers({ [options.attribute || 'content']: defaultContent }),
  ifEditable(withMetaSnippet({ attribute: 'content', ...options })),
  withoutProps('setComponentData'),
  withData,
  renderHoc(options.name),
);

const withMeta = withHeadElement(withMeta$);
const withTitle = withHeadElement(withTitle$);
```

packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -82,24 +86,6 @@ const withMetaStatic = (
...
-      <HelmetComponent {...rest}>
-        {children}
-        <title>{data.content || ''}</title>
-      </HelmetComponent>
-    );
-  }
-  return <HelmetComponent {...rest} />;
-};
-
 const withMetaHtml = (
```
wodenx, line 89: At some point we should refactor this to use the same pattern as withMeta ... but it can probably be descoped, as there's a possibility we may give up on editable sitewide metatags. 
See #472 and #473

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -24,7 +24,11 @@ export type MetaSnippetProps = {
   attribute: string,
 };
 
-export const withMetaSnippet = (data: MetaSnippetProps, next?: Function) => withEditFormSnippet({
+export const withMetaSnippet = (
+  data: MetaSnippetProps,
```
wodenx, line 28: It's confusing that this is called "data".  These are really the options, aren't they? Also, why are the handlers separate positional parameters instead of part of the options.

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -38,27 +42,24 @@ export const withMetaSnippet = (data: MetaSnippetProps, next?: Function) => with
   },
   submitValueHandler: (values: any) => {
     const { name, attribute } = data;
```
wodenx, line 44: The more i think about it, i believe we should simplify and not support a custom attribute.  It's really only partially supported, since in `withMeta$` we hardcode the attribute to "content".

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -76,7 +77,7 @@ const withMetaFormHeader = (headerProps: HeaderProps) => (Component: CT) => {
   return WithFormHeader;
 };
 
-const withMetaForm = (useGetMenuOptions: any, metaFormHeader: HeaderProps) => flowRight(
+const withMetaForm = (useGetMenuOptions: any, metaFormHeader: HeaderProps | undefined) => flowRight(
```
wodenx, line 80: we should provide a default here (use the values currently specified at site level).

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const withMetaSnippet = (
+  data: MetaSnippetProps,
+  nextSubmitHandler?: Function,
+  nextInitialValuesHandler?: Function,
+) => withEditFormSnippet({
+  render: () => {
+    const { name, label, type } = data;
+    const { ComponentFormLabel, ComponentFormText, ComponentFormTextArea } = useFormUI();
+    const Field = type === 'text' ? ComponentFormText : ComponentFormTextArea;
```
wodenx, line 35: A better/more functional way to do this would be to add an optionl 'useFormElement` option and pass in a hook which returns the component to use to render the form.  This could default to ComponentFormText - so here you'd ahve
```
const Field = useFormElement  ? useFormElement() || ComponentFormText;
```
and for descriptoin, have
```
withMetaSnippet({
  ...options,
  useFormElement: () => useFormUI().ComponentFormTextArea,
})
```
That way, we don't have to increase the complexity here as we decide we want to use other form elements for other metatags, and we get rid of this translation from a string/enum to a component.
```
and

File: packages/bodiless-components/src/Meta/Meta.tsx [hunk 22]
```diff
@@ -12,15 +12,62 @@
...
+  withNodeKey(nodeKey),
+  withNode,
+  withNodeDataHandlers({ [options.attribute || 'content']: defaultContent }),
+  ifEditable(withMetaSnippet({ attribute: 'content', ...options })),
+  withoutProps('setComponentData'),
+  withData,
+  withMeta$(options.name),
+);
+
+const withMetaStatic = (
```
wodenx, line 70: I don't like having a completely separate implementation for the static version.  Why not just have
```
const withMetaStatic = (options) => (nodeKey, defaultContent) => flowRight(
  asReadOnly,
  withMeta(options)(nodeKey, defaultContent),
);
```

examples/test-site/src/components/Layout/meta.tsx
```diff
@@ -0,0 +1,72 @@
...
+  withMetaStatic,
+  withMetaHtml,
+  asBodilessHelmet,
+  withMetaForm,
+} from '@bodiless/components';
+import type { MetaFormFieldType } from '@bodiless/components';
+
+const withMetaTitle = withMeta({
+  name: 'title',
+  type: 'text' as MetaFormFieldType.Text,
```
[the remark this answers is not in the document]
dewen: MetaFormFieldType defined as type, can't be used as value here.

packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -16,75 +16,68 @@ import React, { ComponentType as CT, PropsWithChildren } from 'react';
...
-import type { FieldType } from './withMetaForm';
 
 type BaseProps = PropsWithChildren<HelmetProps>;
 type Data = {
   content: string;
 };
 type Props = BaseProps & Data;
 
-type Options = {
+type BasicOptions = {
```
wodenx, line 31: Where is this used?

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -19,37 +19,39 @@ export type HeaderProps = {
...
+    const { name, submitHandler: nextSubmitHandler } = options;
+    const submitValues = { content: values[name] };
     return nextSubmitHandler ? nextSubmitHandler(submitValues) : submitValues;
   },
   initialValueHandler: (values) => {
-    const { name, attribute } = data;
-    const initialValues = values[attribute] ? {
+    const { name, initalValueHandler: nextInitialValuesHandler } = options;
+    const initialValues = values.content ? {
       ...values,
```
wodenx, line 53: Why do we include the original values - won't this give us
```
{
  content: "...content...",
 [name]: "...content...",
}
```

packages/bodiless-components/src/Meta/withMetaForm.tsx
```diff
@@ -77,11 +79,17 @@ const withMetaFormHeader = (headerProps: HeaderProps | undefined) => (Component:
...
 };
 
-const withMetaForm = (useGetMenuOptions: any, metaFormHeader: HeaderProps | undefined) => flowRight(
+const defaultMetaFormHeader = {
+  title: 'SEO Data Management',
+  description: `Enter the page level data used for SEO. 
+  This is metadata needed for SEO that will go in the page header.`,
+};
+
+const withMetaForm = (useGetMenuOptions: any, metaFormHeader?: HeaderProps) => flowRight(
```
wodenx, line 88: Can we not type this as any?

packages/bodiless-components/src/Meta/Meta.tsx
dewen: It's the options for non-editable meta data HOC, like withMetaStatic.

examples/test-site/src/components/Layout/meta.tsx
```diff
@@ -44,12 +44,12 @@ const withMetaPageType = withMeta({
   label: 'Page type',
 });
 
-const withMetaPageBrand = withMetaStatic({
-  name: 'brand',
+const withMetaBrand = withMetaStatic({
+  name: 'bl-brand',
```
wodenx, line 48: @hvanyo please verify this is the correct name of the metatag.

packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -34,7 +34,7 @@ type BasicOptions = {
 
 type Options = {
   label: string;
-  useFormElement?: Function,
+  useFormElement?: () => CT,
```
wodenx, line 37: should this be `CT<any>`

packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -34,7 +34,7 @@ type BasicOptions = {
 
 type Options = {
   label: string;
-  useFormElement?: () => CT,
+  useFormElement?: () => CT<any>,
```
wodenx, line 37: I think this could be typed to return a component which, at least, takes a "field" and "placeholder" prop -- or maybe there's a type applicaable to all informed fields?

packages/bodiless-core/src/asReadOnly.tsx
```diff
@@ -23,7 +23,8 @@ class ReadOnlyContext extends PageEditContext {
 
 const asReadOnly = (Component: ComponentType<any> | string) => (props: any) => {
   const oldContext = useEditContext();
-  const newContext = new ReadOnlyContext(oldContext);
+  // @ts-ignore: root context has no parent.
```
wodenx, line 26: Instead of ignoring this, can we fix the types to handle this?

dewen: Addressed on shared branch:
https://github.com/johnsonandjohnson/Bodiless-JS/pull/485/commits/5aa5144d9b2c6905aa6200eea08a6da63def0e69

packages/bodiless-components/src/Meta/Meta.tsx
dewen: Addressed on shared branch:
https://github.com/johnsonandjohnson/Bodiless-JS/pull/485/commits/5aa5144d9b2c6905aa6200eea08a6da63def0e69

examples/test-site/src/components/Layout/meta.tsx
hvanyo: @wodenx  It is -- we prefix some custom meta-tags with bl-.  
Our thought in future is other components might re-use the meta-tag data set.   At this point, there isn't a use case.  The first use case was page title -> "TITLE | BRAND" but we determined if they wanted to do this now, they would enter in that phrase and or do at site level.

dewen:
PR moved to 
https://github.com/johnsonandjohnson/Bodiless-JS/pull/485

Issue #324, opened 2020-05-11
Component Picker: a single component as a result of searching or filtering takes a  half of the expected width


TatsianaMiraniuk:
## Description 
Component Picker: a single component as a result of searching or filtering takes a half of the expected width

### Steps to reproduce
1. start Edit env. Switch to Edit mode
2. open a page with a flowContainer, e.g. /flow-container/
3. click inside a flowContainer
4. click + in the Editor Menu
5. expand Types and select Single Accordion
### Expected result
Single accordion component is shown. The width of the component picker area isn't changed

### Actual result
Single accordion component is shown but the width of the results area is a half less
![Screen Shot 2020-05-11 at 5 58 25 PM](https://user-images.githubusercontent.com/3224040/81616184-0922ee00-93b1-11ea-8f5e-9b1f66c21c91.png)

6. click 'All'
7. in Type, select 'Rich Text'  => 3 RTE show are expected
![Screen Shot 2020-05-11 at 5 49 16 PM](https://user-images.githubusercontent.com/3224040/81616503-8f3f3480-93b1-11ea-8abb-b1799d9ab541.png)
8. in Rich Text filter, select any, e.g Simple Rich Text
### Actual result
the width of the results area is narrowed
![Screen Shot 2020-05-11 at 5 49 22 PM](https://user-images.githubusercontent.com/3224040/81616501-8ea69e00-93b1-11ea-9c89-e86e458276dc.png)

### Environment
System:
OS: macOS High Sierra 10.13.4
CPU: (4) x64 Intel(R) Core(TM) i7-5650U CPU @ 2.20GHz
Shell: 3.2.57 - /bin/bash
Binaries:
Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
Yarn: 1.17.0 - /usr/local/bin/yarn
npm: 6.14.1 - ~/.nvm/versions/node/v10.15.0/bin/npm
Languages:
Python: 2.7.10 - /usr/bin/python
Browsers:
Chrome: 81.0.4044.138
Safari: 11.1
npmGlobalPackages:
gatsby-cli: 2.8.29

### Affected Version
reproduced in [Master](https://github.com/johnsonandjohnson/Bodiless-JS/commit/e16e2995dd88e0aca7ec1fbd21586358d327accd)

TatsianaMiraniuk:
fixed by https://github.com/johnsonandjohnson/Bodiless-JS/pull/415

PR #415, opened 2020-07-02
Fix (core-ui): Component Picker - the component items are flatten after first opening

### LINKED ISSUE #348: Component Picker: the component items are flatten after first opening
## Description 
Component Picker: the component item are flatten after first opening

### Steps to reproduce
1. open Edit env, Home page
2. switch to Edit mode
3. click in a tout item to get '+' in Editor Menu
4. click +
### Expected result
Screen Shot 2020-05-19 at 4 51 35 PM

### Actual result
Screen Shot 2020-05-19 at 4 50 36 PM

### Note:
the items get to their normal view when some actions are made in Component picker: e.g. if search or filter is used, or if a component is selected and Component Picker is reopened after that.

### Environment
System:
OS: macOS High Sierra 10.13.4
CPU: (4) x64 Intel(R) Core(TM) i7-5650U CPU @ 2.20GHz
Shell: 3.2.57 - /bin/bash
Binaries:
Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
Yarn: 1.17.0 - /usr/local/bin/yarn
npm: 6.14.1 - ~/.nvm/versions/node/v10.15.0/bin/npm
Languages:
Python: 2.7.10 - /usr/bin/python
Browsers:
Chrome: 81.0.4044.138
Safari: 11.1
npmGlobalPackages:
gatsby-cli: 2.8.29

### Affected Version
reproduced in [Master](https://github.com/johnsonandjohnson/Bodiless-JS/commit/455fd7c39be04ebe16ec2236ecb21b4785ca8af1)

konst8:
## Changes
Make the components in the picker to be flexible by height if the horizontal scale is 1.

## Test Instructions
Open component picker, make sure the components are not flattened.

## Related Issues
Fixes #348 #324


packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
```diff
@@ -65,15 +65,17 @@ const ItemList: React.FC<ItemListProps> = props => {
 
   useEffect(() => {
     // Imitate nextTick() functionality to make sure components are fully mounted.
-    setTimeout(() => {
```
beliayeu, line 68: @konst8, did you remove setTimeout on purpose?

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
```diff
@@ -65,15 +65,17 @@ const ItemList: React.FC<ItemListProps> = props => {
...
-      // eslint-disable-next-line no-undef
-      document.querySelectorAll('.bl-outerTransform').forEach(element => {
-        // we have to cast this so that we can use style and offsetHeight
-        const htmlElement = element as HTMLElement;
+    // We are in useEffect so we know document exists
+    // eslint-disable-next-line no-undef
+    document.querySelectorAll('.bl-outerTransform').forEach(element => {
+      // we have to cast this so that we can use style and offsetHeight
+      const htmlElement = element as HTMLElement;
+      if (scale === 1) {
```
beliayeu, line 73: it would be nice to avoid having magic numbers

beliayeu: it would be nice to add a comment explaining why the problem happens only for scale = 1 and does not happen on other scales

beliayeu, line 73: I believe we mitigated the issue, but we did not solve it completely
per my understanding we may still get the issue once the scale is changed but images are not loaded yet by browser
I believe we should either make recalculation after all dependent assets are loaded or come up with another mechanism how to scale items

VancheeZze: I'd also add that component picker most likely renders the components themselves, not pictures of them. So it could be that the client needs some extra time to render those components before the height could be set.

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
konst8: Yes. I guess it was a try to fix the issue, which didn't work apparently.

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
konst8: Added height recalculation on image load.

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
```diff
@@ -65,15 +65,20 @@ const ItemList: React.FC<ItemListProps> = props => {
...
-        htmlElement.style.height = `${htmlElement.offsetHeight / scale}px`;
-      });
+    // We are in useEffect so we know document exists
+    // eslint-disable-next-line no-undef
+    document.querySelectorAll('.bl-outerTransform').forEach(element => {
+      // we have to cast this so that we can use style and offsetHeight
+      const htmlElement = element as HTMLElement;
+      htmlElement.style.height = '100%';
+      htmlElement.style.height = `${htmlElement.offsetHeight / scale}px`;
+      const img = htmlElement.querySelector('img');
```
beliayeu, line 75: consider supporting
* images embedded using picture
* other Embedded content html elements (iframe, video, audio)
* elements containing multiple images

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
beliayeu: @konst8, I think comment "Imitate nextTick() functionality to make sure components are fully mounted." above can be removed since it is related to setTimeout thing

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
konst8: Done. Should now recalculate on load of any asset that has 'src' attribute.

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
```diff
@@ -84,9 +60,7 @@ const ItemList: React.FC<ItemListProps> = props => {
             {Component.title || Component.displayName || 'Untitled'}
           </finalUI.TitleWrapper>
           <div className="bl-outerTransform bl-w-full bl-bg-white">
-            <div
-              style={transformStyle}
-            >
+            <div>
```
VancheeZze, line 63: I would also remove this extra <div>

packages/bodiless-layouts/src/ComponentSelector/ItemListScale.tsx
```diff
@@ -145,14 +139,16 @@ const ItemList: React.FC<ItemListProps> = props => {
...
       </finalUI.ScalingHeader>
-      <finalUI.GridListBoxInner>
-        {elems}
-        {
-          components.length > 50
-            ? <MoreItems />
-            : Fragment
-        }
-      </finalUI.GridListBoxInner>
+      <div style={{ height: '500px' }}>
```
beliayeu, line 142: it would be nice to add a comment why we setting the height to 500px

konst8: @beliayeu Currently the whole picker form has static size:
https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-ui/index.tailwind.css#L54
So, 500 - is the form height considering heading and paddings.

konst8: To my mind, the form still needs some attention in terms of styling. But probably in the scope of another ticket.

PR #419, opened 2020-07-06
feat(gatsby-theme-bodiless): Retry on Error When Saving to the Backend

### LINKED ISSUE #394: Retry on Error When Saving to the Backend
## User Story
As a content editor
I want failed attempts to save data to be retried  
So that I do not lose work.

## Acceptance Criteria
AC1. When an error occurs saving data to the backend, the request is retried until successful
AC2. If a new request to modify the same node is made, and the original request has no succeeded, the original request is abandoned and the new request is retried until successful  
AC3. The contents of a node in the browser (in the client) for which there is a pending retry should not be updated from the data provided by Gatsby

beliayeu:
## Changes
* when an error occurs after saving data to the backend, the request is retried until successful
* retry delay is incrementally increased between subsequent failed requests

## Related Issues
Fixes #394

### DIFF
--- a/packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
+++ b/packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
@@ -27,6 +27,8 @@ enum ItemState {
   Queued,
 }
 
+const DEFAULT_REQUEST_DELAY = 500;
+
 export default class GatsbyMobxStoreItem {
   @observable data = {};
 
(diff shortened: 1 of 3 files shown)


packages/gatsby-theme-bodiless/src/dist/GatsbyMobxStoreItem.ts
```diff
@@ -89,6 +94,12 @@ export default class GatsbyMobxStoreItem {
         this.setState(ItemState.Locked);
         this.setLockTimeout();
         break;
+      case ItemStateEvent.OnRequestError:
+        // incrementally increasing time between each subsequent retry
```
wodenx, line 98: I wonder if we should have a limit here - i mean a maximum delay. Not sure what is reasonable, maybe 5 minutes?

beliayeu: fixed in 492ac84e1ca57fb6cfc04c5a0743528bdd5dff30

PR #420, opened 2020-07-06
[WIP] feat(core): Admin UI: Admin SubMenu

VancheeZze:


packages/gatsby-theme-bodiless/src/dist/Page.tsx
```diff
@@ -49,29 +45,39 @@ const defaultUI: FinalUI = {
...
+  label: 'File',
+  name: 'file',
+}];
+
+const GitMenu = flowRight(
+  withSubmenu({
+    useGetMenuOptions,
+    name: 'File',
+    title: 'File',
+    getSubMenuButtons: getGitButtons,
```
wodenx, line 64: The trouble with this approach is that it requires the component providing the main menu button to know all the submenu buttons which might be added.  I believe this violates AC6 (at least as we discussed it at standup the other day -- it is not a very clear AC).  The idea, I think, is that you should be able to create a submenu context and then any child should be able to add buttons.  This is exactly the same pattern as `withCompoundForm` -- the component which provides the form doesn't need to know anything about the children who contribute fields to the f0rm.

packages/bodiless-core/src/components/ContextSubMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+import type { TMenuOption } from '../PageEditContext/types';
+import withChild from '../withChild';
+import ContextMenu from './ContextMenu';
+import { useUI as useFormUI } from './ContextMenuItem';
+
+type SubMenuOptions = {
+  useGetMenuOptions: any,
+  name: string,
+  title: string,
+  getSubMenuButtons: () => TMenuOption[],
```
wodenx, line 28: As i said in the other comment - we need to decouple the submenu buttons from the submenu.

wodenx: Maybe it would be as simple as this:
```
diff --git a/packages/bodiless-core/src/components/ContextMenu.tsx b/packages/bodiless-core/src/components/ContextMenu.tsx
index f2636112..1313263f 100644
--- a/packages/bodiless-core/src/components/ContextMenu.tsx
+++ b/packages/bodiless-core/src/components/ContextMenu.tsx
@@ -12,7 +12,7 @@
  * limitations under the License.
  */
 
-import React from 'react';
+import React, { ReactElement } from 'react';
 import { getUI as getFormUI } from '../contextMenuForm';
 import { IContextMenuProps as IProps, UI } from '../Types/ContextMenuTypes';
 import { TMenuOption } from '../PageEditContext/types';
@@ -36,14 +36,22 @@ export const getUI = (ui: UI = {}) => ({
   ...ui,
 });
 
-const ContextMenu = (props: IProps) => {
+const ContextMenu: React.FC<IProps> = (props) => {
   if (typeof window === 'undefined') {
     return <></>;
   }
 
-  const { options, ui } = props;
+  const { options, ui, children } = props;
   const { Toolbar } = getUI(ui);
 
+  const childOptions = React.Children.map(children, (child, i) => (
+    // @todo I get some weird type error on the next line
+    React.cloneElement(child as ReactElement, {
+      ui,
+      index: options.length + i,
+    })
+  ));
+
   const elements = options
     // Inject dividers
     .reduce(
@@ -64,7 +72,7 @@ const ContextMenu = (props: IProps) => {
           ui={ui}
         />
       ),
-    );
+    ).concat(<>{childOptions}</>);
 
   if (elements.length > 0) {
     return (
```
and then
```
const useRegisterSubMenuOption = (option: TMenuOption) => useRegisterSnippet({
  id: 'foo', // @todo figure out how to get a unique id.
  render: () => (
    <ContextMenuItem
      option={option}
      key={option.name}
      aria-label={option.name}
      index={0}
      ui={{}}
    />
  ),
  // @todo these should be optional
  initialValues: {},
  submitValues: () => undefined,

})
```
and we'd have to build on withCompoundForm so that instead of always rendering
```
  return (
    <ContextMenuForm {...props} {...formProps}>
      {snippets.map(s => s.render(renderProps))}
    </ContextMenuForm>
  );
```
we could inject a wrapper for the snippets, like
```
  return (
    <ContextMenuForm {...props} {...formProps}>
     <Wrapper>
        {snippets.map(s => s.render(renderProps))}
      <.\/Wrapper>
    </ContextMenuForm>
  );
```
Which, in our case, would be the <ContextSubMenu> component.

packages/bodiless-core/src/withSubMenu.tsx
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @param options Hook to register Admin Menu option and its name.
+ */
+const withSubmenu = <P extends object>(options: SubMenuOptions<P>) => {
+  const formOptions = { hasSubmit: false };
+  const finalOptions = { ...options, formOptions, peer: true };
+
+  return flowRight(
+    withDesign({ Wrapper: addProps({ title: options.name }) }),
+    designable(subMenuComponentsStart),
```
wodenx, line 68: I think we should to move the designable to `withCompoundForm` and just apply the design here, but this is great - we haven't really used the design api in our admin ui yet, but we should start doing it more!

packages/bodiless-core/src/withSubMenu.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * Should be invoked within a component wrapped in `withSubmenu`.
+ *
+ * @param option The TMenuOption object that will be used to render `ContextMenuItem`.
+ */
+const useRegisterSubMenuOption = (option: TMenuOption) => useRegisterSnippet({
+  id: `submenu-${option.name}`,
+  render: () => (
+    <ContextMenuItem
+      option={option}
+      key={option.name}
```
wodenx, line 85: Ah I see - you provide the key here.  That works.

packages/bodiless-core/src/withSubMenu.tsx
```diff
@@ -0,0 +1,96 @@
...
+  Wrapper: SubMenuBody,
+};
+
+/**
+ * HOC to create a Sub Menu which is built on top of the "compound form". Children of this
+ * component can contribute Menu Options to the Sub Menu form.
+ *
+ * @param options Hook to register Admin Menu option and its name.
+ */
+const withSubmenu = <P extends object>(options: SubMenuOptions<P>) => {
```
wodenx, line 62: We should rename this to avoid confusion with the `withSubMenu` in organisms.  Maybe 'withAdminSubmenu'?

packages/bodiless-core/src/components/ContextMenu.tsx
```diff
@@ -62,9 +76,19 @@ const ContextMenu = (props: IProps) => {
           key={option.name}
           aria-label={option.name}
           ui={ui}
+          setRenderForm={allowTooltips ? undefined : setRenderForm}
         />
       ),
-    );
+    ).concat(<React.Fragment key="child-options">{childOptions}</React.Fragment>);
```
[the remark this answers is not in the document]
VancheeZze: We just need to have a `key` here for the fragment and `<>...</>` won't work in this case

packages/bodiless-core/src/components/ContextMenu.tsx
```diff
@@ -36,14 +36,28 @@ export const getUI = (ui: UI = {}) => ({
...
+  const {
+    options,
+    ui,
+    allowTooltips = true,
+    children,
+  } = props;
   const { Toolbar } = getUI(ui);
 
+  const childOptions = React.Children.map(children, (child, i) => (
+    React.cloneElement(child as ReactElement, {
```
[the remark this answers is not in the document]
VancheeZze: Yep, we moved the `key` prop to the child.

packages/bodiless-core/src/contextMenuForm.tsx
```diff
@@ -47,32 +47,28 @@ export type Options<D> = {
...
-  closeForm,
-  onClose,
-  ui,
-  submitValues = () => undefined,
-  initialValues = {} as D,
-  hasSubmit = true,
-  children = () => <></>,
-  ...rest
-}: Props<D>) => {
+export const ContextMenuForm = <D extends object>(options: ContextMenuPropsType<D>) => {
```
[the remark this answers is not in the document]
VancheeZze: Yes, I was tracing some prop in type and just updated the naming and usage a little. I think at some point I had this type used somewhere else.

packages/bodiless-core/src/components/ContextMenuItem.tsx
```diff
@@ -40,8 +40,14 @@ export const useUI = () => {
...
-const ContextMenuItem = ({ option, index, ui }: IProps) => {
-  const [renderForm, setRenderForm] = useState<(props:FormProps) => JSX.Element>();
+const ContextMenuItem = (props: IProps) => {
+  const {
+    option,
+    index,
+    ui,
+    setRenderForm,
+  } = props;
+  const [renderForm, setRenderFormChild] = useState<(props:ContextMenuFormProps) => JSX.Element>();
```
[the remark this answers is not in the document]
VancheeZze: Done

packages/bodiless-core/src/Types/ContextMenuTypes.tsx
```diff
@@ -63,10 +63,19 @@ export type IContextMenuProps = {
   // onDispose?: (wasSubmitted: boolean) => void;
   ui?: UI;
   isPositionToggled?: boolean;
+  allowTooltips?: boolean;
```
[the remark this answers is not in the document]
VancheeZze: Updated to `renderInTooltip` - `setParentRenderForm={renderInTooltip ? undefined : setRenderForm}`

packages/bodiless-core/src/withSubMenu.tsx
VancheeZze: Moved `designable` into `withCompoundForm`.

packages/bodiless-core/src/withSubMenu.tsx
VancheeZze: Named it `withContextSubMenu`.

packages/bodiless-core/src/components/ContextMenu.tsx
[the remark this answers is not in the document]
wodenx: why do we need a key for the fragment?  Why do we need the fragment at all?  Why not just the children, each of which has its own key?

packages/bodiless-core/src/components/ContextMenuItem.tsx
```diff
@@ -45,9 +45,9 @@ const ContextMenuItem = (props: IProps) => {
     option,
     index,
     ui,
-    setRenderForm,
+    setParentRenderForm, // setParentRenderForm setRenderForm
```
wodenx, line 48: sorry this isn't exactly what i mean - i meant pass the prop as `setRenderForm` but rename it to `setParentRenderForm` when you destructure.

PR #422, opened 2020-07-07
feat(core-ui): add a warning element for user warning messages. 

### LINKED ISSUE #416: Warnings/Error Treatment
## User Story

As a Content Editor, I want more noticeable warnings/errors, so that I am aware that some action needs to be taken. 

### Acceptance Criteria

AC1:

When the Content Editor takes an action that has a warning/error, then there is a warning Icon with a message in the form panel. 
This treatment will not be applied to operations that use overlay methods improperly and generate errors in invalid ways outside of the form panel such as: Add Page, Save, Revert or use alert("error")

AC2:

The icons should be a reusable element in the context menu form UI. 

### Implementation Notes

 Tech Note:

For any case when we have warnings/errors the treatment of the warning icon is added to the existing message. 
The icon is named "Warning" 
This should be consistent on all warnings and errors in forms

awm086:
## Changes
- Add and use a warning form element for in the pull changes panel when conflict exists

## Test Instructions
- Create a conflict that cannot be pulled.
- Click pull changes.
- Expect to see a warning as described in the issue #416 

## Related Issues
- Closes #416

### DIFF
--- a/packages/gatsby-theme-bodiless/src/dist/RemoteChanges.tsx
+++ b/packages/gatsby-theme-bodiless/src/dist/RemoteChanges.tsx
@@ -112,7 +112,7 @@ const ChangeContent = ({ status, masterStatus, errorMessage } : ContentProps) =>
       if (masterStatus === ChangeState.CanNotBePulled) {
         return (
           <ComponentFormWarning>
-            Alert! There are changes on production which cannot be merged without developer support.
+            There are changes on production which cannot be merged from the UI.
           </ComponentFormWarning>
         );
       }
@@ -127,8 +127,7 @@ const ChangeContent = ({ status, masterStatus, errorMessage } : ContentProps) =>
               initiate.
             </div>
             <ComponentFormWarning>
-              Alert! There are changes on production which cannot be merged
-              without developer support.
+              There are changes on production which cannot be merged from the UI.
             </ComponentFormWarning>
           </>
         );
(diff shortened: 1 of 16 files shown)


packages/gatsby-theme-bodiless/src/dist/RemoteChanges.tsx
```diff
@@ -107,32 +106,44 @@ type ContentProps = {
...
       );
+
     case ChangeState.CanNotBePulled:
       return (
-        <>Upstream changes are available but cannot be fetched via the UI.</>
+        <ComponentFormWarning>
+          Upstream changes are available but cannot be fetched via the UI.
+        </ComponentFormWarning>
       );
     case ChangeState.Errored:
```
hvanyo, line 148: The error message should be treated with the same form.

hvanyo:
@awm086  The intent of the story was to put warning icons different warnings/errors that happen throughout system not just on pull.

i.e. 
- new page creation: https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/gatsby-theme-bodiless/src/dist/NewPageProvider.tsx#L53
- https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-core/src/PageEditContext/index.tsx#L271

konst8:
@awm086 
The feature works as expected on 'pull' operation when there are conflicts between remote branches (production and upstream):


Should we consider Heidi's comment that "The intent of the story was to put warning icons different warnings/errors that happen throughout system not just on pull"? What would be the additional test cases then?

Also, the travis build is failed for some reason. Please have a look. Thanks.

PR #440, opened 2020-07-14
feat(fclasses): Allow adding/removing classes conditionally via fclasses

### LINKED ISSUE #432: Allow adding.removing classes conditionally via fclasses.
## User Story
As a site builder
I want to be able to use the design API to toggle styling based on component state
So that I can implement dynamic designs.

## Background

The current bodiless way to do this is to use a flowToggle and put a different component in place depending on state, eg:
```
const Component = designable(..)({ Title, Body }) => {
  const [on, setOn] = useState(false);
  return (
    <Wrapper on={on} >
      <Body />
    </Wrapper>
  );
};
const useToggle = props => props.on;

const asColoredWrapper = withDesign(
  Wrapper: flow(
    ifToggledOn(useToggle){removeClasses(blue), addClasses(red))),
    ifToggledOff(useToggle)(removeClasses('red'), addClasses('blue')),
  ),
);
```
The problem here is that when this toggle is triggered, the entire component tree under the Wrapper will be destroyed and re-rendered bc the Wrapper component itself is being swapped.  The problem is illustrated in [this code sandbox](https://codesandbox.io/s/dreamy-tereshkova-zc50d).

As an alternative, we want an API which allows
```
const asColoredWrapper = withDesign(
  Wrapper: flow(
    addClassesIf(useToggle)('red'),
    removeClassesIf(useToggle)('blue'),
    addClasesIfNot(useToggle)('blue'),
     removeClassesIfNot(useToggle)('red')
  ),
);
```
In wihch case the actual component rendered will remain the same (I think), an donly the props will be toggled.

In scope of this we should also add `addPropsIf`, `addPropsIfNot`, `removePropsIf` and `removePropsIfNot` which can be used where something other than classes need to be modified.

## Acceptance Criteria

- The fclasses api has a toggle styling based on component state.
- The documentation is updated accordingly.
- Test cases exist or updated for changes.

beliayeu:
## Changes
* added helpful functions that allows to add/remove classes conditionally via fclasses
* removing chaining support from addClasses/removeClasses
* added an example to test-site that demonstrates how to add classes to a component based on its state
* updated unit tests
* updated doc

## Test Instructions
* review provided api
* check the example on test site

## Related Issues

Fixes #432

BREAKING CHANGE: chaining support is removed from addClasses and removeClasses. If your code contains chained addClasses/removeClasses, you need to replace the chaining with flow/flowIf or any other helpers that provides functional composition.
For example, when you have in your code
```
addClasses('classA').removeClasses('classB')
```
you need to change it to
```
flow(
  addClasses('classA'),
  removeClasses('classB'),
)
```


packages/fclasses/src/FClasses.tsx
```diff
@@ -35,33 +35,54 @@ type Classable = {
...
 
-const modifyClasses = (operation: 'add' | 'remove') => (classes?: Classes) => {
+export type Condition = () => boolean;
+const alwaysTrueCondition = () => true;
+
+const modifyClassesIf = (operation: 'add' | 'remove') => (condition: Condition) => (classes?: Classes) => {
   const hoc = <P extends StylableProps>(Component: ComponentType<P> | string) => {
     const ModifyClasses: FC<P> = ({ fClasses, ...rest }) => {
-      const newFClasses = {
+      const newFClasses = condition() ? {
```
wodenx, line 44: I think we should pass the props to the condition, as we do for flow toggle.

packages/fclasses/src/FClasses.tsx
```diff
@@ -35,33 +35,54 @@ type Classable = {
...
       return (
         <Component fClasses={newFClasses} {...rest as P} />
       );
     };
     ModifyClasses.displayName = `${capitalize(operation)}Classes`;
     return ModifyClasses;
   };
-  hoc.addClasses = (classes$1?: Classes) => flow(hoc as HOC, modifyClasses('add')(classes$1) as HOC);
-  hoc.removeClasses = (classes$1?: Classes) => flow(modifyClasses('remove')(classes$1) as HOC, hoc as HOC);
+  hoc.addClassesIf = (condition$1: Condition) => (classes$1?: Classes) => flow(hoc as HOC, modifyClassesIf('add')(condition$1)(classes$1) as HOC);
```
wodenx, line 56: I'd like to discuss with @e2thex whether it's worth preserving the ability to chain here.  the flow pattern is now ubiquitous, and maybe this is an unnecessary complication.

packages/fclasses/src/FClasses.tsx
```diff
@@ -35,33 +35,54 @@ type Classable = {
...
   return hoc;
 };
 
+/**
+ * Allows to add classes to a component conditionally.
+ *
+ * @param condition A function that is evaluated to determine whether classes should be added.
+ * @returns HOC that can be used for adding classes to a component
+ */
+const addClassesIf = modifyClassesIf('add');
```
wodenx, line 70: Do we need 'addClassesIfNot` and 'removeClassesIfNot`? or should we provide an invert function, so you can do
```
addClassesIf(condition)(classes),
removeClassesIf(condition)(classes$),
addClassesIf(invert(condition))(lasses$),
removeClassesIf(invert(condition))(classes),
```
I think I like the latter option better.

packages/fclasses/src/FClasses.tsx
beliayeu: makes sense, fixed in 7df1a176f17c27b9c70b50dd075e6cb56f4a411e

packages/fclasses/src/FClasses.tsx
beliayeu: I think it would be nice to export this functionality.
I do not have strong arguments, but I like the former option better (addClassesIfNot/removeClassesIfNot)

e2thex: I do not think we need an invert function or the Not functions we have !

packages/fclasses/src/FClasses.tsx
e2thex: I think changing did not have depth anyways so I think it is better to remove it.

packages/fclasses/src/FClasses.tsx
e2thex: So https://lodash.com/docs/4.17.15#negate can do the job

packages/bodiless-core-ui/src/ContextWrapper.tsx
```diff
@@ -28,9 +32,12 @@ type VariantProps = {
 
 const ContextWrapperDiv = flow(
   withoutProps<VariantProps>(['isActive']),
-  addClasses('bl-border bl-border-transparent').flow,
+  addClasses('bl-border bl-border-transparent'),
   flowIf(hasProp('isActive'))(
```
wodenx, line 36: Should we refactor this to use the new API - instead of flowIf, use addClassesIf and removeClassesIf?

beliayeu: fixed in 063eefb4b388430183e22c203ab5aacc4a4b1e21

PR #455, opened 2020-07-17
fix(test-site): Reusable Logo

hvanyo:
## Changes

- Change Logo to reusable between deskheader and mobile burger menu
- Allows site logo to be specified in gatsbyconfig.js and updating the logo there changes it both places 
- Renamed ExampleHelmet -> SiteHelmet because example didn't make sense.
- Split out design for header/footer/logo to token.ts
- Renamed index.jsx -> index.tsx

## Test Instructions
* On Static, Edit & Edit preview -- verify header, logo, footer and burger-menu all operate as before. 
There is slight change in padding for burger menu to give space around the burger icon.


examples/test-site/src/components/Layout/header.tsx
```diff
@@ -42,28 +43,24 @@ export type Props = {
 const headerComponents:HeaderComponents = {
   Wrapper: Div,
   Container: Div,
-  SiteReturn: Div,
   Menu: ResponsiveMenu,
+  SiteLogoReturn: Logo,
```
VancheeZze, line 47: Here might be a typescript error, I think `HeaderComponents ` does not have `SiteLogoReturn` but instead has `Logo`.

examples/test-site/src/components/Layout/header.tsx
```diff
@@ -34,6 +34,7 @@ type HeaderComponents = {
   Container: ComponentType<any>,
   SiteReturn: ComponentType<any>,
```
VancheeZze, line 35: I think `SiteReturn` is no longer used? We could probably remove it from this type.

packages/bodiless-organisms/src/components/BurgerMenu/BurgerMenuClean.tsx
```diff
@@ -39,9 +39,11 @@ const burgerMenuComponents: BurgerMenuComponents = {
   Header: Div,
 };
 
-type Props = DesignableComponentsProps<BurgerMenuComponents> & HTMLProps<HTMLElement>;
+export type Props = {
+  siteLogo: string,
+} & DesignableComponentsProps<BurgerMenuComponents> & HTMLProps<HTMLElement>;
```
VancheeZze, line 44: We already have this type defined and exported in `examples/test-site/src/components/Menus/index.tsx`. We could probably rename it into something like `MenuProps` and reuse it here as well. This way if we decide to extend it again in the future we won't have to worry about updating both places

examples/test-site/src/components/Layout/header.tsx
hvanyo: Addressed

examples/test-site/src/components/Layout/header.tsx
hvanyo: Addressed

packages/bodiless-organisms/src/components/BurgerMenu/BurgerMenuClean.tsx
hvanyo: Addressed

examples/test-site/src/components/Layout/logo.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+const Logo: FC<Props> = ({ siteLogo, components }) => {
+  const {
+    SiteReturn,
+    SiteLogo,
+  } = components;
+
+  return (
+    <SiteReturn>
+      <Link to="/">
```
wodenx, line 47: why isn't this also part of the design?

examples/test-site/src/components/Layout/logo.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+const asLogo = flow(
+  designable(logoComponents),
+  withDesign({
+    SiteReturn: addClasses('py-4 px-2'),
+    SiteLogo: addClasses('h-16'),
+  }),
+);
+
+export default asLogo(Logo);
```
wodenx, line 61: Seems like we should export the clean logo and the token, not the combination.

examples/test-site/src/components/Layout/header.tsx
```diff
@@ -42,28 +42,24 @@ export type Props = {
...
 
   return (
     <Wrapper>
       <Container>
-        <SiteReturn>
-          <Link to="/">
-            <img src={siteLogo} className="h-16" alt="Return To Home" />
-          </Link>
-        </SiteReturn>
+        <SiteLogoReturn siteLogo={siteLogo} />
```
wodenx, line 59: Instead of passing the logo through here, consider adding it as a prop via an HOC which gets it from wherever we get it (a static query, i think?)

packages/bodiless-organisms/src/components/BurgerMenu/BurgerMenuClean.tsx
```diff
@@ -51,13 +53,14 @@ const BurgerMenuBase: FC<Props> = ({ components, ...rest }) => {
   return (
     <Wrapper>
       <Slide>
-        <Header />
+        <Header siteLogo={siteLogo} />
```
wodenx, line 56: Same here - instead of passing the logo url in as a prop, why not just attatch it to the Logo component which is used here.  That way you completely decouple the BurgerMenu from any dependency on the site logo.

examples/test-site/src/components/Layout/logo.tsx
hvanyo: Addressed

examples/test-site/src/components/Layout/logo.tsx
hvanyo: Addressed

packages/bodiless-organisms/src/components/BurgerMenu/BurgerMenuClean.tsx
hvanyo: Removed

examples/test-site/src/components/Layout/header.tsx
hvanyo: I think this is close -- but not sure if was suppose to define it outside of LogoClean definition.  Its still within.

Natalia1809:
Testing is completed. Works as expected.

asiarheyeu:
Approving on behalf of @Natalia1809

PR #474, opened 2020-07-23
feat(documentation): API Doc for Site builder

### LINKED ISSUE #389: Integrate typedoc into documentation build
## User Story
As a developer
I want comprehensive API documentation for all public Bodiless APIs
So that I understand how to use them

## Acceptance Criteria
AC1. When I build any bodiless package, API documentation is automatically produced via typedoc, and published in markdown format along with the rest of the package docs.

AC2. When I build documentation for a site, the API docs for all bodiless packages are included in the final documentation hierarchy (Development > API).

AC3. The Bodiless Documentation package illustrates the above

beliayeu:
## Changes
* added api documentation generation on bodiless package build
* added publishing of package api documentation in markdown format along with the rest of the package docs
* added api documentation to the documentation generated for a site
* added a toggle to opt-in and opt-out inclusion of api documentation to site documentation. disabled by default. in order to enable api documentation, `BODILESS_DOCS_API_ENABLED` env variable should be set to '1'
* disabled api documentation section on test site

## Notes
* typedoc is leveraged for api documentation generation
* package level build script is extended with api doc generation
* extended bl-docs-build script with logic to copy api documentation to the doc server
* extended bl-docs-build script with adding a link to api documentation during runtime

## Related Issues

Closes #389


examples/test-site/package.json
```diff
@@ -61,6 +61,7 @@
     "build:watch": "run-p -r dev-frontend dev-backend",
     "build:lib": "gatsby build",
     "build:doc": "bl-docs-build",
+    "build:doc-dev": "../../node_modules/.bin/ts-node --dir ../../packages/bodiless-documentation src/blDocsBuildDev",
```
wodenx, line 64: Is there a reason we separate this?  I also don't think we should have the test site refer to the node_modules from the mmonorepo.

packages/bodiless-components/package.json
```diff
@@ -21,6 +21,7 @@
   "scripts": {
     "build": "npm run copy && tsc -p ./tsconfig.json",
     "build:watch": "npm run build -- --watch",
+    "build:api-doc": "typedoc --out doc/api --mode file",
```
wodenx, line 24: Do we need to maintain this in each package rather than running it from monorepo root with lerna exec?

packages/bodiless-documentation/src/blApiDocsBuild.ts
```diff
@@ -0,0 +1,78 @@
...
+
+type Props = {
+  copier: Copier;
+  targetDocPath: string;
+};
+
+const generateIndexFile = (baseTargetPath: string, packages: string[]) => {
+  const lines = packages.map(pkg => `* [${pkg}](/Development/API/${pkg}/globals)`);
+  const targetPath = pathUtil.join(baseTargetPath, 'Development', 'API', 'README.md');
+  fs.writeFileSync(targetPath, lines.join('\n'));
```
wodenx, line 54: Is there a reason we're doing this synchronously?

examples/test-site/package.json
beliayeu: this line is added for debugging purposes, will remove it before merging the pr

packages/bodiless-components/package.json
beliayeu: yep, I thought about adding just one npm script to root package.json and use lerna exec similar to what we have for packing but the problem is that not all our packages expose api, so we need to have a way to how to mark a public package if api should be generated for it

packages/bodiless-documentation/src/blApiDocsBuild.ts
beliayeu: there is no special reason
chose using writeFileSync since we do not have a lot of lines and I believe it is more readably then async version

VancheeZze:
I would suggest we add an example of the `BODILESS_DOCS_API_ENABLED ` to the `.env.site` files at `test-site` and `starter`.

Issue #465, opened 2020-07-22
Can not reorder components immediately after adding them.


konst8:
## Description
STR
1. In a flow container, add a few components.
1. Grab the drag handle and try reorder

ER
Can reorder the components added.

AR
Cannot reorder the components added.
If you click Edit to go to the "review state (preview) state" and click edit again THEN you can reorder the components.

PR #475, opened 2020-07-24
Fix (flow-container): Can not reorder components immediately after adding them

konst8:
## Changes
- Unify structure of empty flow container

## Test Instructions
- Create a new page
- Add a few flow containter items
- Try to reorder them right after adding without reloading the page

## Related Issues
Fixes https://github.com/johnsonandjohnson/Bodiless-JS/issues/465


packages/bodiless-layouts/src/FlowContainer/SortableContainer.tsx
```diff
@@ -43,11 +43,13 @@ const SortableListWrapper = SortableContainer(
         const { FlowContainerEmpty } = getUI(ui);
         const context = useEditContext();
         const activeClassName = context.isActive ? 'bl-border-orange-400' : 'hover:bl-border-orange-400';
-
+        const classNames = `bl-flex bl-w-full bl-justify-center bl-flex-wrap bl-py-grid-3 ${activeClassName}`;
         return (
-          <FlowContainerEmpty {...rest} className={`bl-flex bl-justify-center bl-flex-wrap bl-py-grid-3 ${activeClassName}`} {...useContextActivator()}>
-            Empty FlowContainer
-          </FlowContainerEmpty>
+          <section {...rest} {...useContextActivator()}>
```
beliayeu, line 48: @konst8, please add root cause of the problem to the ticket description
also, since the fix is to wrap flow container with children and flow container without children with the same parent (section element), consider moving the parent element to a separate component

konst8: Refactored, to make it more clear and consistent. Wasn't able to define the exact root cause, though.

PR #483, opened 2020-07-28
chore(components): examples how to set default alt for an image

beliayeu:
## Changes
* allowed to set default alt for an image
* provided an example how to set alt using withImagePlaceholder hoc
* provided an example how to set alt using prop


hvanyo:
Awesome -- closing example PR for now.

Issue #450, opened 2020-07-16
Implement Interface for Managing Meta Data


schristodoulou:
## Description
As a Content Editor, I want the ability to manage meta data via the interface, so that I do not have to bring in a developer to manage the content. 

https://jhdfdc.axshare.com/#g=1&p=generic_ui

*Form Title:* 

SEO Data Management  

**Form Description Text:**

Enter the page level data used for SEO. This is metadata needed for SEO that will go in the head of the script.  

**Form fields:**
|*Title*|*Field*|*Placeholder text to appear in field*|
|Title|Text Field|rec 30-65 char|
|Description|Long Text Field|rec <160 char|
|Page Type|Text Field| |

Notes:
- No validation 
- No scrolling of the form because it's only 3 pieces 
- Keep in mind that this form may be combined with another form such as GTM meta data at a later time 
- Long text field should be 3 lines high
  

 
## Acceptance Criteria
AC1: The Content Edit UI Main Menu includes an "SEO" Icon and label 

AC2: Clicking on the "SEO" Icon opens a form panel (form.png) that has a:
- Form Title: "SEO Data Management"
- Form Description: "Enter the page level data used for SEO. This is metadata needed for SEO that will go in the page header."
- Form Fields
  - Title, Description and Page Type
  - Text box fields
- The checkmark and X work as expected same as the other panel windows work.

AC3: The data is used and stored from the meta data, the metadata storage structure stays the same

AC4: There are standard set of metatags for title, description and page type which can be added to the page, along with the associated form with little to no code duplication 

AC5: Default placeholders match the text in the description and the values of the fields are empty values. 

## Attachments
<img title="ZoomAnnotationContentWnd.png" src="https://user-images.githubusercontent.com/59970745/87707795-b29aad00-c76f-11ea-91e2-f389e90feaa7.png" width="200" />

dewen:
Question regarding <title> tag rendering: If the value is empty string, should the title tag get rendered?

@schristodoulou @hvanyo ^^

hvanyo:
@dewen  prefer not but that is current behavior.  We gapped in the last meta tag implementation.  If title or other meta are empty and then don't render it, that is preferred behavior.  So if you address it now we close the gap and if you don't we leave gap open.

dewen:
I added title empty check and hide <title> tag if content is empty. But Helmet might have logic to render a default one if no user title overrides.

Issue #391, opened 2020-06-22
Notify of error saving to backend.


wodenx:
## User Story
As a content editor
I want to be informed when there are errors saving data 
So that I do not lose work.

## Acceptance Criteria
AC1. When an error occurs saving data to the back-end, a notification to this effect is displayed
AC2. When an error occurs saving data to the backend, the request is queued for retry.
AC3. The retry queue can never contain more than one request to modify the same node.
AC4. When all queued retries have completed successfully, the notification is removed.

Issue #277, opened 2020-04-14
Admin SubMenu 


hvanyo:
## User Story

- AS A Content Editor
- I WANT submenus 
- SO THAT left admin isn't cluttered and functionality is grouped 

### Acceptance Criteria

AC1.
GIVEN the user is on the edit UI
Then there is a "File" icon on the main menu

AC2.
When the user clicks on the "File" icon 
THEN there is a panel window with a Title and the icon options for Pull, Push, History, Revert (correct order & names)
AND the sub menu panel window includes the "x" icon for closing the window
AND does not include the checkmark icon for accepting interaction. 

AC3.
WHEN the user clicks on the Pull, Push, Revert, History or close icons from the panel window (sub menu)
THEN existing functionality behaves as expected.

AC4. 
Submenu content opens in the same panel window
AND
Closing a submenu items returns back to the parent item(s)

AC5. 
Automated tests are updated to reflect above criteria. 

AC6. 
The "sub menu" in the Page Panel window, that allows for a selection between different functional options, is a reusable component

### Implementation Notes

- Scope of this story includes have a main menu icon that opens a sub menu of existing icons
- The Changeset text is not in this scope and will be part of another story

### References

- Wireframes: https://jhdfdc.axshare.com/#g=1&p=generic_ui&id=sxj4x9
- Relates to Clone Existing Page #70
![Aries](https://user-images.githubusercontent.com/59970745/83162523-dc244a00-a0d7-11ea-923f-c707c158132f.jpg)

schristodoulou:
Scope of this ticket is only 1-level of submenus.

PR #485, opened 2020-07-28
Feat(core, component, ui, gatsby-theme-bodiless): Implement meta data, admin menu and backend error notifications

dewen:
## Changes
- Implement Admin Sub Menu
- Implement SEO Management Form
- Implement Backend error notifications
- Implement Structured Children
- Updated cypress tests

## Related Issues
Closes:  #450, #277, #391


packages/bodiless-components/src/Meta/Meta.tsx
```diff
@@ -34,7 +34,7 @@ type BasicOptions = {
 
 type Options = {
   label: string;
-  useFormElement?: () => CT<any>,
+  useFormElement?: () => CT<HTMLProps<HTMLElement>>,
```
wodenx, line 37: It would be better if we typed this with  at least the props that we are passing it ("field")

wodenx:
@dewen The meta ui part of this looks good.  Now just waiting for the admin menu and we can move it to testing.

PR #487, opened 2020-07-28
feat(gatsby-theme-bodiless): notify on backend error

### LINKED ISSUE #391: Notify of error saving to backend.

beliayeu:
## Changes
* a notification is displayed for the user when an error occurs saving data to the back-end
* the notification is removed when all queued retries are completed

## Notes
* alternative approach is to manage notifications on node level. wip pr - https://github.com/beliayeu/Bodiless-JS/pull/9

## Related Issues
Closes #391


packages/bodiless-core/src/NotificationProvider.tsx
```diff
@@ -50,33 +53,52 @@ const NotifyContext = React.createContext<NotifyContextType>({
...
 
+class NotificationStore {
+  @observable private notificationsMap = new Map<string, NotificationProviderItem[]>();
+
+  @computed get hasNotifications() {
+    return this.notifications.length > 0;
+  }
+
+  @computed get notifications() {
+    return Array.prototype.concat.apply([], Array.from(this.notificationsMap.values()));
```
wodenx, line 64: Why not just return Array.from(...)?

packages/bodiless-core/src/NotificationProvider.tsx
```diff
@@ -35,13 +36,15 @@ type Notifier = (owner: string, notifications: Notification[]) => void;
...
   notifications: Notification[],
+  hasNotifications: () => boolean,
 };
 type NotifyContextType = {
   notify: Notifier,
 };
 
 const NotificationContext = React.createContext<NotificationsContextType>({
   notifications: [],
+  hasNotifications: () => false,
```
wodenx, line 47: Do we need this?  Why not just observe the notifications array?

packages/bodiless-core/src/NotificationProvider.tsx
beliayeu: map's value is array, so we have a collection of arrays, we need to flatten it and remove empty arrays.

packages/bodiless-core/src/NotificationProvider.tsx
beliayeu: there are 2 reasons:
1. if we pass notifications (array) to icon and isActive props of getMenuOptions, then we do not get the corresponding ContextMenuItem rerendered on changes to notifications
2. useNotificationButton is only triggered when notifications become empty and when notifications are changed from empty to non-empty (it will not be triggered when there is existing notification(s) and a new notification is added)

PR #490, opened 2020-07-29
chore: Release v0.0.52

dewen:


VancheeZze:
@dewen @hvanyo Shouldn't we also update root `package.json` and `package-lock.json`?

hvanyo:
> @dewen @hvanyo Shouldn't we also update root `package.json` and `package-lock.json`?

@VancheeZze  good catch!

PR #519, opened 2020-08-20
feat(core): Improved Context Menu API

### LINKED ISSUE #523: Improve Context Menu API
## User Story
As a developer or end user
I want a more robust API for context menu items
To improve performance and reliability

## ACs
AC1. Re-renders of context menu items and subscribers to edit context are minimized
AC2. All menu items update their state automatically when the component providing them re-renderss

wodenx:
## Changes
Major refactor of context menu to optimize re-renders and ensure menu options update automatically when the component providing them re-renders.  Also resolves issues with stale form data when json files change while the context is currently active.
- Menu options are stored in an observable map in the page edit store.  They are updated onEffect whenever a context provider re-renders, or when a component registering peer menu options rerenders.  This only triggers re-renders of the context menu when the set of menu options changes, and only triggers re-renders of context menu items when their properties change.
- Menu option getters are invariant (with the options themselves stored in a ref).  This prevents child components from re-rendering when a parent changes its menu options.
- Menu option providers are optimized:
  - Where possible, they provide invariant menu options and rely on mobx to update the items
   - In general, they provide a fixed set of options, and use `isHidden` to control visibility
   - Menu options are memoized to ensure that any callbacks only change when necessary.
- Resolves an issue in peer context registration which was causing peer contexts which were provided by components which rendered conditionally not to be deleted when those components were removed.
- Removes the `stopPropagation` from context activators, and uses a better mechanism to ensure outer contexts don't activate.
- Simplifies `withEditButton` and `withEditFormSnippet` and eliminates some code duplication.
- global menu option provides are separated into individual components to cheapen re-renders when those menu options change.

## TO-DO
- replace `renderForm` key with `useRenderForm` in EditButtonOptions and remove unnecessary props.  This will allow the handler provided by withEditButton to be memoized depending only on renderForm.
- documentation updates.

## Related Issues
Closes #523 

## BREAKING CHANGE: 
  - The API for withMenuOptions has changed.  Instead of providing a `useGetMenuOptions` hook which returns a `getMenuOptions` callback to simply `useMenuOptions`, which returns the options array itself.
- The `useMenuOptions` key is no longer available in `EditButtonOptions` since you can now use peer contexts to add additional menu options to the existing context.


packages/bodiless-core/src/PageEditContext/Store.tsx
```diff
@@ -0,0 +1,178 @@
...
+    }
+    const map = this.optionMap.get(context.id);
+    const keys = new Set();
+    [context, ...context.peerContexts].forEach(c => {
+      // Update all items in the map
+      c.getMenuOptions().forEach(op => {
+        keys.add(op.name);
+        const existing = map!.get(op.name);
+        if (existing) {
+          Object.assign(existing, op);
```
beliayeu, line 113: my concern is
when there is a prop that does not exist in `op` but exists in `existing` then this prop won't be deleted from map item

packages/bodiless-components/src/List/asEditableList.tsx
```diff
@@ -12,50 +12,42 @@
...
 
-  const asHandler = (action: Function) => () => {
-    action();
-    context.refresh();
-  };
-
-  return () => {
-    const options = [];
-    options.push({
+  const menuOptions = useMemo(() => ([
```
beliayeu, line 30: do we need to wrap a list of options with useMemo before passing to withMenuOptions?

beliayeu (approved):

beliayeu (approved):

TatsianaMiraniuk:
Hi @wodenx ,

please see issue_1:

The expected item isn't shown after filtering by clicking on a radiobox label. The item is shown only if a radiobox is clicked (EDIT env Edit mode only) (/filter-page/, /products/)
Steps to reproduce on /filter-page/
1. add a few items in Filter panel
2. click a label of any item in Filter panel
Actual Result:
'Selected Tags' doesn't reflect that - the selected tag isn't shown. The radiobox is not set as selected
3. click on the radiobox 
Actual Result:
'Selected Tags' now has the expected selected tag. But the radiobox is still not set as selected.
4. click on the label
Actual Result:
The radiobox is now set as selected.

TatsianaMiraniuk:
@wodenx 

please see issue_2:
flow-container:  the page gets blank after adding/deleting a component into a flow-container. There is an error in a console : Rendered more hooks than during the previous render.
RT-Products_Error_local
RT-Products_Error_pl sh

TatsianaMiraniuk:
@wodenx ,

please see **issue_3**:

flow-container: There is no possibility to add more than one component into a flow-container (NO 'Add' icon is shown in Editor Menu after adding first component):

_Steps to reproduce:_
1. open any page with a flow-container, e.g. /flow-container/
2. click inside 'Empty FlowContainer'
3. Click Add+ icon in Editor Menu
4. Click on any component 
5* refresh the page (due to the issue_2)
6.  click on the component to get the flow-container icons in Editor Menu 

_Expected Result:_
Add+ icon is available and working altogether with Swap and Delete

_Actual Results:_
No Add+ is available. Only Swap and Delete are shown.

FlowContainer_NoAddIcon
)

beliayeu (approved):

beliayeu (dismissed):

TatsianaMiraniuk:
> @wodenx
> 
> please see **issue_2**:
> flow-container: the page gets blank after adding/deleting a component into a flow-container. There is an error in a console : Rendered more hooks than during the previous render.
> 
> _Steps to reproduce:_
> 
> 1. open any page with a flow-container, e.g. /flow-container/
> 2. click inside 'Empty FlowContainer'
> 3. Click Add+ icon in Editor Menu
> 4. Click on any component
> 
> _Expected Results:_
> The component is added successfully.
> _Actual Results:_
> The page gets white, there is an error in a console.
> 
> RT-Products_Error_local
> RT-Products_Error_pl sh

Confirmed that the bug is fixed and no longer reproducible.

TatsianaMiraniuk:
Hi @wodenx ,

please take a look at **issue_4**.

Edit Env crashes sporadically (if it ran locally)
or getting 502 error in console  (if it ran in platform.sh)
while working with Main Menu - there are no specific steps. The error was caught when the new items (top and submenu) were added; when they were clicked to get Local Context Menu, or hovered. (Videos are available in the RT spreadsheet.)
Please see screenshots:
MainManuCrashes
MainMenu_502

TatsianaMiraniuk (approved):

hvanyo (approved):

beliayeu (approved):

hvanyo (approved):

Issue #441, opened 2020-07-14
Image files should be uploaded to path per page


hvanyo:
## User Story

As a site builder I would like uploaded images to be uploaded in file folder specific for that page for organization so that they do not accidentally overwrite images from another page.

### Acceptance Criteria

AC1. As image are uploaded with nodekey at page level, they are uploaded to static folder path that is unique to the page and nodekey of image. 

AC2. As image are uploaded with nodekey at site collection level, they are uploaded to static folder path that is unique  to the page and nodekey of image.

### Implementation Notes

Technical notes:
* To isolate images to that page use page path (or UUID) as unique path
* Unique folder path is something like: static/images/pagepath/nodekey/filename where
** pagepath is the page alias
**  If nodekey is long or issues with $ in folder structure, then possibly hash.

* Migration path is out of scope.  Will need to manually fix cases on demo/penaten/example test site.

PR #527, opened 2020-08-26
feat(backend, components): Image files should be uploaded to path per page

VancheeZze:
## Changes
- Images are uploaded to a static folder path that is unique to the page and hash of image `nodePath`.
- New Cypress helper `isImageVisible(imageXpath)` that can be used to test if the image was successfully loaded and its `src` is not 404.
## Test Instructions
- Upload images from different pages and confirm image path matching `/images/pages/{pagepath}/{hash}/{filename}`
- Images uploaded from the home page should be saved to `/images/site/{hash}/{filename}`

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/441


packages/bodiless-components/src/Image.tsx
```diff
@@ -70,6 +70,20 @@ const defaultImagePickerUI = {
   UploadStatus: ({ statusText }: UploadStatusProps) => <div>{statusText}</div>,
 };
 
+const generateHash = (str: string) => {
```
beliayeu, line 73: @VancheeZze 
1. lets discuss if we want to have hashing generation logic on client side or server side
2. discuss how to avoid the function duplication in cypress
3. add some details in comment regarding chosen hashing algorithm

packages/bodiless-components/src/Image.tsx
```diff
@@ -99,16 +113,23 @@ function DropZonePlugin({ formApi, targetFieldName, ui }: {
     return () => null;
   });
   const onDrop = useCallback(acceptedFiles => {
+    const pagePath = window.location.pathname.replace(/^\/|\/$/g, '');
```
beliayeu, line 116: it would be nice to
1. add a comment regarding what we are doing here
2. probably move the logic to get fileDist in a separate function. uploaded filename and page path will be params

beliayeu, line 116: will it work if upload multiple images with the same name on a page? I think we should take into consideration nodeKey of the image.

packages/bodiless-components/src/Image.tsx
VancheeZze: Updated as discussed

packages/bodiless-components/src/Image.tsx
VancheeZze: Updated as discussed

VancheeZze: Updated as discussed

packages/bodiless-backend/src/fileHelper.js
```diff
@@ -0,0 +1,48 @@
...
+});
+
+const generateHash = (str) => crypto.createHash('md5').update(str).digest('hex');
+
+const copyAllFiles = (files, pagePath, nodePath) => {
+  const allFiles = [];
+  Object.keys(files).forEach(key => allFiles.push(files[key]));
+
+  return Promise.all(allFiles.map(file => {
+    const distFolderPath = path.join(backendStaticPath, 'images', pagePath, generateHash(nodePath));
```
beliayeu, line 36: I think we should account pagePath only for Page nodes. Site nodes (shared between pages) should go to images path directly. Given that we need to consider avoiding conflict between site images and home page page level images

cypress/integration/smokeTest/pdp_spec.js
```diff
@@ -81,15 +81,14 @@ describe('PDP (Product Details Page) smoke tests', function () {
...
       .click();
-    const fileName = imageName
-    cy.fixture(fileName).then(fileContent => {
-      cy.get('input[type=file]').upload({ fileContent, fileName, mimeType: "image/jpeg" });
+    const imagePath = `${imagesFolderPath}/${imageName}`;
+    cy.fixture(imagePath).then(fileContent => {
+      cy.get('input[type=file]').upload({ fileContent, fileName: imageName, mimeType: "image/jpeg" });
     })
-    cy.wait(3000);
+    cy.wait(1000);
```
beliayeu, line 88: any reasons we are decreasing this?

cypress/integration/smokeTest/pdp_spec.js
```diff
@@ -81,15 +81,14 @@ describe('PDP (Product Details Page) smoke tests', function () {
...
     cy.xpath(checkmarkIconImageFormXpath)
       .click();
-    cy.xpath(imagePlaceholderXpath)
-      .should('have.attr', 'src', '/' + imageName);
```
beliayeu, line 92: am I right that we are removing the case to test what image src is displayed?

VancheeZze: It is now replaced with `cy.isImageVisible(imagePlaceholderXpath)`. As discussed we now don't just test the `src` of the image as it is generated on the backend. Instead, we check that image is loaded and it is not 404.

cypress/integration/smokeTest/pdp_spec.js
VancheeZze: No strong reasons -- changed it during some tests and never updated back. Although it is good enough to wait only 1 sec, we can return it back to 3s

packages/bodiless-backend/src/fileHelper.js
VancheeZze: Addressed to check for the global node path on backend and ignore `pagePath` if node is global.

cypress/integration/smokeTest/pdp_spec.js
VancheeZze: Returned to 3s

beliayeu (approved):

TatsianaMiraniuk (approved):

PR #531, opened 2020-09-01
feat(gatsby-theme-bodiless): Image generation on upload

### LINKED ISSUE #477: Image generation on upload
## Description
As a Site builder/editor, I want the tool (Gasby Image) to generate scaled default images, so to reduce overhead of resizing images manually.

Leverage work done in this SPIKE:

Technical Note:
- The assumption is that the images would already be optimized by the agency.  
- First iteration scope is so that a Developer can do this, Out of scope and will be a later issue is creating the form for the content editor experience 
- Proceed with Option 2 from spike

## Acceptance Criteria
There exists a HOC that provides the following (AC1 - AC5):
AC1
 Loads the optimal size of image for each device size and screen resolution with defaults:  360,834,1024 

AC2
 Holds the image position while loading so your page doesn’t jump around as images load

AC3
 Uses the “blur-up” effect i.e. it loads a tiny version of the image to show while the full image is loading

AC4
 Alternatively provides a “traced placeholder” SVG of the image.

AC5
 Lazy loads images which reduces bandwidth and speeds the initial load time

AC6
 An API is available that allows you to add a specific preset to an image.
 
AC7 
 In edit mode, when the image is not yet processed, the uploaded image will display.

AC8 
 Extend the images test page to demostrate test cases.

## Resources:
Demo: https://using-gatsby-image.gatsbyjs.org/traced-svg/

beliayeu:
## Changes
* exported asGatsbyImage hoc from @bodiless/gatsby-theme-bodiless that can be used by site builders for rendering images using gatsby-image component
* exported a bunch of default image presets from @bodiless/gatsby-theme-bodiless
* extended test-site images page with examples of rendering image using gatsby-image
* removed gatsby-transformer-code from @bodiless/gatsby-theme-bodiless and replaced the functionality provided by the plugin with custom code so that get more flexibility

## Related Issues

Closes #477


packages/gatsby-theme-bodiless/create-node.js
```diff
@@ -0,0 +1,352 @@
...
+  webp: true,
+  tif: true,
+  tiff: true,
+};
+
+const srcSetBreakpoints = [
+  360,
+  834,
+  1024,
+];
```
VancheeZze, line 93: Is there an opportunity to retrieve the breakpoints from the tailwind config?

VancheeZze (dismissed):

packages/gatsby-theme-bodiless/bin/bl-clear-logs
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+
```
VancheeZze, line 2: Do we need copyright text added here below the shebang string?

packages/gatsby-theme-bodiless/bin/bl-validate-logs
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env node
+
```
VancheeZze, line 2: Copyright text here as well

VancheeZze (approved):

packages/gatsby-theme-bodiless/create-node.js
beliayeu: per discussion in standup that is will be done in a follow-up ticket

packages/gatsby-theme-bodiless/bin/bl-clear-logs
beliayeu: fixed in 949e6312a70d124b218c96d895b9b510f3753216

packages/gatsby-theme-bodiless/bin/bl-validate-logs
beliayeu: fixed in 949e6312a70d124b218c96d895b9b510f3753216

VancheeZze (approved):

dewen (approved):

VancheeZze (approved):

Issue #411, opened 2020-06-30
Only have close on forms with no user interactions


hvanyo:
## User Story

As a Content Editor, When I interact with informational only admin functions (ie. PUll button), then I only need to see the Checkmark or Close because it is the same functionality. 

### Acceptance Criteria

AC1

When a model window has no interaction then only the X appears for closing the panel window and the checkmark is not available (removed from current see screenshot)

### Implementation Notes

- pull
- commit
- Alert
- history
- Revert
- Push

PR #564, opened 2020-09-21
feat(core): Remove submit button from non-interactive forms

awm086:
## Changes
- Remove submit button from non-interactive forms.

## Test Instructions
- see https://github.com/johnsonandjohnson/Bodiless-JS/issues/411

## Related Issues
- Fixes https://github.com/johnsonandjohnson/Bodiless-JS/issues/411


hvanyo (dismissed):
Pseudo approver for Ivan -- reviewed cypress changes.

cypress/integration/smokeTest/fileSubmenu_spec.js
```diff
@@ -159,7 +156,6 @@ describe('File Submenu Smoke Tests', function () {
     cy.xpath(historyFormItems)
       .its('length')
       .should('be.gt', 3)
-    cy.xpath(historyFormSubmitButton)
```
TatsianaMiraniuk, line 162: @awm086 ,

I would suggest to check that the submit button doesn't exist anymore (won't appear in future as a result of  regression), so instead of removing that line, let's change it to
cy.xpath(historyFormSubmitButton)
   .should('not.exist')

cypress/integration/smokeTest/fileSubmenu_spec.js
```diff
@@ -169,7 +165,6 @@ describe('File Submenu Smoke Tests', function () {
       .click()
     cy.xpath(revertFormTitle)
     cy.xpath(revertFormDescription)
-    cy.xpath(revertFormSubmitButton)
```
TatsianaMiraniuk, line 172: @awm086 ,

same as above. let's check that it doesn't exist:
.should('not.exist')

cypress/integration/smokeTest/fileSubmenu_spec.js
```diff
@@ -37,16 +37,13 @@ describe('File Submenu Smoke Tests', function () {
   // History Form
   const historyFormTitle = '//*[@aria-label="Context Submenu Form"]//h3[text()="Latest Commits"]'
   const historyFormItems = '//*[@aria-label="Context Submenu Form"]//input[@type="radio"][@name="commits"]'
-  const historyFormSubmitButton = '//*[@aria-label="Context Submenu Form"]//*[@aria-label="Submit"]'
```
TatsianaMiraniuk, line 40: should be kept for the check described below

cypress/integration/smokeTest/fileSubmenu_spec.js
```diff
@@ -37,16 +37,13 @@ describe('File Submenu Smoke Tests', function () {
...
   const historyFormCloseButton = '//*[@aria-label="Context Submenu Form"]//*[@aria-label="Cancel"]'
 
   // Revert Form
   const revertFormTitle = '//*[@aria-label="Context Submenu Form"]//h3[text()="Revert to saved"]'
   const revertFormDescription = '//*[@aria-label="Context Submenu Form"]//label[text()="Discard local changes"]'
-  const revertFormSubmitButton = '//*[@aria-label="Context Submenu Form"]//*[@aria-label="Submit"]'
```
TatsianaMiraniuk, line 46: should be kept for the check described below

TatsianaMiraniuk (changes requested):

TatsianaMiraniuk (approved):

VancheeZze (approved):

PR #581, opened 2020-10-01
test(organisms): fix an image uploading check

### LINKED ISSUE #579: Image Uploading: better handling of the feature is needed in Cypress tests.
touts_spec.js passes but in fact an image is not uploaded in a tout component.

That brought a point that we need to return to the original check when an image file name of an element is verified, instead of the recently implemented one  (isImageVisible) that checks that an image is not 0x0 as it's true for the default placeholder and thus it shows false-positive result.

touts_spec.js and pdp_spec.js are expected to be updated.

TatsianaMiraniuk:
## Changes  
- added back a check for 'src' value of an uploaded image, considering a new image path 

## Related Issues
 Closes #579


hvanyo (approved):

cypress/integration/smokeTest/pdp_spec.js
```diff
@@ -15,37 +15,40 @@
...
+  const accordionDirectionsBodyPlaceholderXpath = '//*[@data-accordion-element="accordion"][@aria-label="Directions"]//*[@data-accordion-element="accordion-body"]//*[text()="Enter Product Information"]';
+  const bvTextXpath = '//*[@data-product-element="ratings-summary"][text()="Please hover and click to enter Bazaarvoice Product External ID: "]';
+  const editBVIconXpath = '//*[@aria-label="Local Context Menu"]/*[@aria-label="Settings"]';
+  const closeBVFormXpath = '//*[@aria-label="Context Menu Settings Form"]//*[@aria-label="Cancel"]';
+  const imagePlaceholderXpath = '//*[@data-product-element="image"]';
+  const imageIconXpath = '//*[@role="toolbar" and @aria-label="Local Context Menu"]//*[@aria-label="Image"]';
+  const checkmarkIconImageFormXpath = '//form[@aria-label="Context Menu Image Form"]//button[@aria-label="Submit"]';
+  const flexboxXpath = '//*[@data-product-element="flow-container"]';
+  const addComponentIconXpath = '//button[@aria-label="Add"]';
+  var imagePathRegex = new RegExp("images\/pages" + pdpPagePath + "\/[a-zA-Z0-9]+\/" + imageName, "");
```
VancheeZze, line 50: Not a big deal, but we might want to keep it consistent and declare it with `const` as other variables. I don't see we are changing it anywhere.

cypress/integration/smokeTest/touts_spec.js
```diff
@@ -15,37 +15,40 @@
...
+  const urlFieldXpath = '//form[@aria-label="Context Menu Link Form"]//input[@id="link-href"]';
+  const linkIconXpath = '//*[@role="toolbar" and @aria-label="Local Context Menu"]//*[@aria-label="Link"]';
+  const checkmarkIconLinkFormXpath = '//form[@aria-label="Context Menu Link Form"]//button[@aria-label="Submit"]';
+  const altFieldXpath = '//form[@aria-label="Context Menu Image Form"]//input[@id="image-alt"]';
+  const checkmarkIconImageFormXpath = '//form[@aria-label="Context Menu Image Form"]//button[@aria-label="Submit"]';
+  const imageIconXpath = '//*[@role="toolbar" and @aria-label="Local Context Menu"]//*[@aria-label="Image"]';
+  const ctaButtonXpath = '//*[@id="tout-horizontal"]//a[@data-tout-element="link"]';
+  const imageLinkXpath = '//div[@id="tout-horizontal"]//a[@data-tout-element="image-link"]';
+  var imageOrigPathRegex = new RegExp("images\/pages"+ pagePath + "[a-zA-Z0-9]+\/" + imageNameOriginal, "");
+  var imageUpdPathRegex = new RegExp("images\/pages" + pagePath + "[a-zA-Z0-9]+\/" + imageNameUpdated, "");
```
VancheeZze, line 46: Same here, I'd keep it as `const` here as well.

cypress/integration/smokeTest/pdp_spec.js
TatsianaMiraniuk: thanks! done

cypress/integration/smokeTest/touts_spec.js
TatsianaMiraniuk: done

VancheeZze (approved):

konst8 (approved):

PR #612, opened 2020-10-21
fix(richtext): [GAPS] RTE editor doesn't allow for inline files/components (aka inline CTA styled buttons) #481

### LINKED ISSUE #481: [GAPS] RTE editor doesn't allow for inline files/components (aka inline CTA styled buttons)
## Description
**Issue:**

If you style a cta to be inline element (*Inline elements* are those which only occupy the space bounded by the tags defining the element) the placeholder doesn't hold the space, so there is no way to click into the CTA to enter text.   

**Workaround in keeping with RTE:**
- Place a min-width on the element that is large enough to hold the placeholder but keep it the smallest size so it doesn't affect short content (short content if entered and less than min-width would have odd space at end and not appear centered.) 

**Workaround in switching to Basic Editable:**
- An inline cta – when you replace AsEditorSimple with asEditable – you can click anywhere in the inline cta placeholder will work
- The styling (aka button outline) around the placeholder also has correct dimensions of the placeholder, 
- We can remove the workaround of min-width issue with use of asEditable.
- BONUS:  Editable gives blue focus box. where

marcopagliarulo:
## Changes
Implemented the PlaceholderPlugin provided by slate-react-placeholder in order to set the inline components width to the Placeholder's width

## Test Instructions

1. open Edit env, /touts
2. switch to Edit mode
3. Edit CTA on third Tout
4. Edit Title on fourth Tout
5. Edit Title and CTA on fifth Tout

### Expected Results:
Once removed the text and rendered the placeholder, the inline element keep the placeholder width, allowing to edit again the element.

## Related Issues
Fixes #481


hvanyo:
@marcopagliarulo  

Can you also remove this workaround on the test site menu where it also occurs:
https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/examples/test-site/src/components/Menus/token.tsx#L29
and I believe its copied to starter-site as well. :) 

and then purge the comments in the tailwind in test site & starter site:
/examples/test-site/tailwind.config.js
/examples/startersite/tailwind.config.js
and remove the minWidth overrides of 1 & 5

    // Min-width is set for https://github.com/johnsonandjohnson/Bodiless-JS/issues/481
    // We need a min width here for the asToutOverlayTitle and asToutOverlayCta
    minWidth: {
      1: '1rem',
      5: '5rem',

marcopagliarulo:
The current fix create a visual regression, once focused the editor, the cursor is placed at the right side.
The visual regression can be resolved adding floating to right the placeholder, but this fix has as side effect the bug 
https://github.com/ianstormtaylor/slate/issues/1237
#1234 is already fixed on slate, but not yet available in any release

packages/bodiless-richtext/src/core/SlateEditor.tsx
```diff
@@ -53,6 +56,27 @@ const withSlateEditor = <P extends object> (Component:ComponentType<P>) => (prop
     return change;
   };
 
+  const pluginWithPlaceholder = (placeholder) ? useMemo(() => [...plugins,
+    [
+      {
+        queries: {
+          // @ts-ignore
+          isEmpty: (editor) => editor.value.document.text === '',
```
dewen, line 64: for what reason we set ts-ignore here? can we add the reason?

marcopagliarulo: it was an error, I added it as temporary but then forgot to remove it adding the parameter type. I am fixed it, please check

packages/bodiless-richtext/src/core/SlateEditor.tsx
```diff
@@ -15,11 +15,14 @@
...
   useRef,
   useState,
+  useMemo,
   Fragment,
   ComponentType,
 } from 'react';
 import { Editor, Value } from 'slate';
-import { Editor as ReactEditor, EditorProps } from 'slate-react';
+import { Editor as ReactEditor, EditorProps, Plugin } from 'slate-react';
+// @ts-ignore
```
dewen, line 24: same here, what are we ignoring?

packages/bodiless-richtext/package.json
```diff
@@ -46,6 +46,7 @@
     "shelljs": "^0.8.4",
     "slate": "^0.44.12",
     "slate-react": "^0.21.18",
+    "slate-react-placeholder": "^0.1.20",
```
dewen, line 49: The version for slate v0.44.12 is v0.1.15, shall we use that one?

packages/bodiless-richtext/src/@types/slate-react-placeholder.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'slate-react-placeholder' {
```
dewen, line 1: not sure if we need a types with scope sign folder for this declaration.

packages/bodiless-richtext/package.json
marcopagliarulo: slate-react-placeholder is used by slate-react, slate react 0.21.* has as dependency slate-react-placeholder ^0.1.20

packages/bodiless-richtext/src/core/SlateEditor.tsx
marcopagliarulo: Running the tests the @types are not recognised, instead the types are recognised on build. Without the //ts-ignore, the test fail

dewen (approved):

dewen (approved):

hvanyo (approved):

Issue #593, opened 2020-10-07
Updates to the Main Menu Links


hvanyo:
## Description
As a Content Editor, I want clear UX when interacting with Main Menu Links, so that I am aware when there is a link value or not. 
## Acceptance Criteria
~~AC1. "Remove Link" button has a deactivated state (grayed out) and does not disappear~~

AC2. When a link value is removed by clicking on the "Remove Link" button, the field returns to the initial state with a placeholder value 

[This part will now be done in #499]
~~AC. The link icon label is replaced to "Edit" when there is a link, 

AND

"Add" when there is not a link~~

NOTES:
Note that the `rc-menu` will be deprecated and `rc-menu` links are out of scope.

Issue #499, opened 2020-08-05
Context Menu Stacks by Content Layers with Basic Groupings 


schristodoulou:
## Description
As a Content Editor, I want to easily associate available content actions by area in which I am working in, so that I can quickly access the functions I need to work with. 
- Referring to when the Context Menu appears; The context menu layering

https://jhdfdc.axshare.com/#g=1&p=aries&id=3ct389 

## Acceptance Criteria
AC1:

The buttons of the context menu are divided into named sections

AC2:

The section titles are the Component Name, or Container (for the flow container) 

AC3. 

The layers are ordered horizontally with inner most (on top/to the left) to outer most based on the component they are applied to. 

 

## Attachments
<img title="Aries.jpg" src="https://user-images.githubusercontent.com/59970745/89414606-ae233d80-d6f8-11ea-92de-1c0a353b4b0c.jpg" width="200" />

PR #672, opened 2020-11-09
feat(components, ui): Updates to the Main Menu Links

VancheeZze:
## Changes
- When a link value is removed, the field returns to the initial state.

## Test Instructions
See https://github.com/johnsonandjohnson/Bodiless-JS/issues/593

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/593


dewen (dismissed):

TatsianaMiraniuk:
@VancheeZze ,

please see **issue_1:**
Item label is removed if 'Remove Link' is clicked.

**Steps to Reproduce:**
1. add an item in MainMenu
2. add some text in i, e.g. 'item 1'
3. open Link form for the item
4. add a url in the field
5. click V
6. Reopen Link form for same item
7.  Click 'Remove Link'

**Expected Results:**
Link field _only_ is cleared and gets its initial placeholder '/link'

**Actual Result:**
the whole item get cleared - Item label is removed too.

TatsianaMiraniuk:
@schristodoulou , @wodenx , @VancheeZze 

From the original story #593 

```
[This part will now be done in #499]
~~AC. The link icon label is replaced to "Edit" when there is a link, 

AND

"Add" when there is not a link~~
```

However it does not work as stated after implementing #499 - for Main Menu it always says 'Edit' even if a link wasn't set yet.

Question is - is that needed to be fixed in the current story?

VancheeZze:
> @schristodoulou , @wodenx , @VancheeZze
> 
> From the original story #593
> 
> ```
> [This part will now be done in #499]
> ~~AC. The link icon label is replaced to "Edit" when there is a link, 
> 
> AND
> 
> "Add" when there is not a link~~
> ```
> 
> However it does not work as stated after implementing #499 - for Main Menu it always says 'Edit' even if a link wasn't set yet.
> 
> Question is - is that needed to be fixed in the current story?

I think since this issue is reproduced in the ond deprecated `rc-menu` we should not spend much time on trying to fix this issue.

cc: @TatsianaMiraniuk

VancheeZze:
> @VancheeZze ,
> 
> please see **issue_1:**
> Item label is removed if 'Remove Link' is clicked.
> 
> **Steps to Reproduce:**
> 
> 1. add an item in MainMenu
> 2. add some text in i, e.g. 'item 1'
> 3. open Link form for the item
> 4. add a url in the field
> 5. click V
> 6. Reopen Link form for same item
> 7. Click 'Remove Link'
> 
> **Expected Results:**
> Link field _only_ is cleared and gets its initial placeholder '/link'
> 
> **Actual Result:**
> the whole item get cleared - Item label is removed too.

This issue is Fixed.

dewen (approved):

dewen (approved):

TatsianaMiraniuk (approved):

Issue #621, opened 2020-10-23
Retrofit product filters to use new list API


wodenx:
## User Story
As a developer
I want a single API for lists
So that I don't get confused

## Acceptance Criteria
AC1. The product filters use the new list api
AC2. Local context menu for product filters match spec
AC3. The sublists for product categories are not optional.

schristodoulou:
![image](https://user-images.githubusercontent.com/59970745/97330877-24929000-184f-11eb-923a-d28f64124d1f.png)
![image](https://user-images.githubusercontent.com/59970745/97330937-34aa6f80-184f-11eb-950f-fdb1343105c5.png)

PR #685, opened 2020-11-12
feat(fclasses): Design Keys

### LINKED ISSUE #660: Provide design key attributes in markup
## User Story
As a site builder
I want html elements generated by Bodiless to be marked with design keys
So I can more easily determine how to style them.

## Acceptance Criteria

AC1. HTML elements rendered by designable elements are marked with a custom attribute indicating which design key may be used to add props or classes.

AC2: The design key attribute displays both the design key and the name of the designable component, if available.

AC3. The attribute that is added is valid HTML (has a data prefix) 

AC4. The behavior can be controlled by an environment variable, which is enabled by default on the edit/review environment and is disabled by default on the preview environment. 

Example:

image

## Notes
- Should we add this attribute in all modes (edit/preview/static)?
- Link to [POC branch](https://github.com/wodenx/Bodiless-JS/tree/feat/design-key-attribute)

wodenx:
## Changes
- Add "data-bl-design-key" attribute to designable elements.
- Control presence of this attribute based on "withShowDesighKeys" hoc
- Enable/disable in gatsby-theme-bodiless

## Related Issues

Closes #660


VancheeZze (approved):

hvanyo (commented):
- Search: ?:SearchWrapper, ?:SearchInput, ?:SearchButton
- Iframe: on page: /iframe/ uses Embed:Wrapper, can we change it to use Iframe?
- Filters: /products under Filter:TagList there is unidentified list ?:Item
- RTE: /rte doesn't produce any bl-design-keys
- Image: /image doesn't produce any bl-design-keys as its not designable..  leave as is?
- Editable: are not producing bl-design-keys as its not designable..  leave as is?
- Links: are not producing bl-design-keys as its not designable..  leave as is?

wodenx:
> * Search: ?:SearchWrapper, ?:SearchInput, ?:SearchButton
> * Iframe: on page: /iframe/ uses Embed:Wrapper, can we change it to use Iframe?
> * Filters: /products under Filter:TagList there is unidentified list ?:Item
I'll fix these.  Search wasn't merged yet when I did this

> * RTE: /rte doesn't produce any bl-design-keys
I think we should leave this as is.  Not sure it will be common to apply design to components after they have been put into RTE.

> * Image: /image doesn't produce any bl-design-keys as its not designable..  leave as is?
> * Editable: are not producing bl-design-keys as its not designable..  leave as is?
> * Links: are not producing bl-design-keys as its not designable..  leave as is?
What w9uld you expect for these?  If they are not designable, how could they have design keys?

-

hvanyo (commented):
in .env.site can you this to serve as documentation (both test & starter site)
```
# Enable to help debug in static mode
# BODILESS_DEBUG='1'
```

wodenx:
> Search: ?:SearchWrapper, ?:SearchInput, ?:SearchButton

Fixed

> Iframe: on page: /iframe/ uses Embed:Wrapper, can we change it to use Iframe?

This is because it is the "Embed" component which is designable.  It can't be overridden, and it would be confusing to do so I think.

> Filters: /products under Filter:TagList there is unidentified list ?:Item

This is because the product filters are still using the deprecated list api. Should be fixed by #621 

> This case isn't working.
> set .env.site BODILESS_DEBUG='1', run npm build and serve -- verify bl-design-key's are in static mode

Fixed and added example to .env.site in test-site and starter.

hvanyo (approved):

VancheeZze (approved):

PR #707, opened 2020-11-19
fix: moved package css compilation and purge to site level

### LINKED ISSUE #683: Better handling for tailwind classes referenced in reusable packages
As a site builder I want css purge functionality to happen and not present with issues relying on css precedence for multiples of same classes or order of media queries or order of include of stylesheets. 

## ACS

AC1: bodiless packages (@bodiless/layouts and @bodiless/organisms) have tailwindo configuration and export their tailwind configuration. They do not generate tailwind css files and thus remove purging at package level 
AC2: the tailwind styling from bodiless packages in imported at gatsby-theme-bodiless level, merged.
AC3: There is ability to whitelist packages at site level to prevent purging of packages.  The whitelist list of bodiless packages and is exported from gatsby-theme-bodiless. 
AC4: purge of css happens at site level only 
AC5: feature is implemented on both test site & starter site.
AC6: pr to microlax that applies the changes and process @imodium global brand package.

## Background

There are several bodiless packages which refer to tailwind classes
- bodiless-layouts uses them to set widths
- bodiless-organisms uses them eg for tout layout tokens
- etc...

Further, we expect brand (or other) design libraries to make heavy use of tailwind classes.

Our current methodology is that a package does not export its tailwind config, instead it generates (and purges) whatever css it needs.

Here are the use-cases:
1. package refers to a default tailwind class, and expects a consumer to define the precise meaning of this class (this should be the case for most bodiless packages) 
2. package refers to a *custom* tailwind class, and expects a consumer to define the precise mening of this class (example?)
3. package refers to a  default or custom tailwind class and wants to define the precise meaning of that class (should be rarely true for bodiless packages, but normal for brand design packages)

Specific Proplems
- If a package does not export css for a class, and the class is not used on the site level, it will be purged.
- If a package exports css for a class, it is difficult for site to override that css.  For many classes (which are determinate in nature) this is not a problem, but for some it can be a problem.  In particular, responsive classes (like w-full). Consider:
package:
```
.w-full { width: "100%"; }
```
site:
```
@media (min-width: 768px) {
  .w-full { width: 100% }
}
```

beliayeu:
## Changes
* changed the approach how package level css is compiled and purged. moved the compilation and purge from package level into site level

## Related Issues

Fixes #683


packages/gatsby-theme-bodiless/tailwind.base.config.js
```diff
@@ -0,0 +1,37 @@
...
+      return null;
+    }
+  })
+  .filter(Boolean)
+  .map(indexFilePath => path.dirname(path.join(indexFilePath)))
+  .map(moduleRootPath => moduleRootPath.concat(purgePackageRule));
+
+module.exports = {
+  purge,
+  theme: {
```
wodenx, line 24: Do we want to define this here?  Or should we define this in organisms (or wherever it is used) and import from there?
Or -- should we just document that this should be defined at site level>

packages/gatsby-theme-bodiless/tailwind.base.config.js
```diff
@@ -0,0 +1,37 @@
+const path = require('path');
+
+const whitelistedModules = [
```
wodenx, line 3: Not sure this is the right place - why not at site level?

packages/gatsby-theme-bodiless/tailwind.base.config.js
```diff
@@ -0,0 +1,37 @@
...
+  purge,
+  theme: {
+    aspectRatio: { // defaults to {}
+      none: 0,
+      square: [1, 1],
+      '16/9': [16, 9],
+      '4/3': [4, 3],
+      '21/9': [21, 9],
+    },
+  },
```
VancheeZze, line 32: The `aspectRatio` is also defined in `@bodiless/organisms` tailwind config which is included here anyway. Do we want to remove it from here? Or maybe from `@bodiless/organisms`

wodenx (approved):

packages/gatsby-theme-bodiless/tailwind.base.config.js
beliayeu: changed per discussion at standup

packages/gatsby-theme-bodiless/tailwind.base.config.js
beliayeu: no changes here per agreements at standup

packages/gatsby-theme-bodiless/tailwind.base.config.js
beliayeu: fixed

wodenx (approved):

hvanyo (dismissed):

hvanyo (approved):

wodenx (approved):

Issue #681, opened 2020-11-12
Migrate CI to Github Actions


wodenx:
## User Story
As a bodiless developer
I want reliable CI for Bodiless-JS
So that I can work efficiently

## Acceptance Criteria
AC1. When a pull request is issued to the bodiless-js repository on GitHub, Github actions are triggered to perform the following:
- clean install
- build
- jest unit tests (npm run test)
- lint
- cypress tests

AC2. The "master" branch of the bodiless-js repository is renamed "main", and set as the default branch, with the same branch restrictions as the current "master" branch.

AC3. All existing pull requests to "master" are reissued to "main"

AC4. The CI actions from AC1 are also executed on a push to the new "main" branch.

AC5. If a pull request build fails, the pull request cannot be merged. 

## Notes
- [node build template](https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-nodejs)
- [cypress action](https://www.cypress.io/blog/2019/11/20/drastically-simplify-your-testing-with-cypress-github-action/)

PR #724, opened 2020-11-26
Chore: github action workflow

awm086:
### Changes:
- Add github actions to build, lint and test.

### Example Runs:
- https://github.com/awm086/Bodiless-JS/runs/1457298278
- https://github.com/awm086/Bodiless-JS/actions/runs/384589225

### Test Instructions: 
- Ensure Current master build and pass all tests.
- Merge into the main branch.
- Issue PR to the main branch.

### Related
Related #681


.github/workflows/build-test-lint-simple.yml
```diff
@@ -1,44 +1,62 @@
...
-        env:
-          # make sure every Cypress install prints minimal information
-          CI: 1
+      - name: Print Git Version
+        run: git --version
+
+      - name: Setup git user
         run: |
-          npm ci
+          git config --global user.name "$(git --no-pager log --format=format:'%an' -n 1)"
```
dewen, line 39: this is to get last commit user as global git user?

awm086: Yes

awm086: It has no impact on anything except get conflict test which pulls global user. If we could remove the need for global user in the tests then we could remove this as well.

dewen: I mean we don't need real user here, could be fake one:
```
      - name: Setup git user
        run: |
          git config --global user.name 'gituser'
          git config --global user.email 'gituser@example.com'
```

.github/workflows/build-test-lint-simple.yml
```diff
@@ -1,44 +1,62 @@
 name: Build, Test, Lint for Pull Requests
+env:
+  CI: true
+  FORCE_COLOR: 1
 
 on:
+  push:
```
dewen, line 7: Do we need this push event? There's no direct push to main branch, I think.

If in certain cases, direct push enabled, we might not need to perform the build either.

VancheeZze (approved):

dewen (approved):

hvanyo (approved):

PR #726, opened 2020-11-27
feat(migration-tools): create json files for scraped images

### LINKED ISSUE #641: Create json files for scraped images
## User Story
As a site migrator
I want data for the images on a page automatically written to the page directory
So I can more quickly locate and select images

## Acceptance Criteria

AC1. When the migration tool downloads an image for a page, it places a json file containing image src and alt text (in the schema defined by the bodiless image component) into the /src/data/pages/... directory for the corresponding page.
- The src is the relative url of the downloaded image.
- The alt text is the alt text from the original site.

AC2. The filename of the json file is created by prefixing the name (without extension) of the original image with a string which can be specified as a configuration option, eg:
```
imageDataPrefix: '___assets$`
```

AC3. The above behavior can be enabled/disabled by a configuration option.

dewen:
## Related Issues

fixes #641


packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -99,33 +145,40 @@ export default class Downloader {
...
+            return Promise.reject(
+              new Error(`Resource ${target} has been excluded from download.`),
+            );
           }
           ensureDirectoryExistence(targetPath);
-
-          return req.pipe(fs.createWriteStream(targetPath))
-            .on('finish', resolve)
-            .on('error', err => reject(new Error(`error on streaming ${resource}. ${err}.`)));
+          return req.pipe(fs.createWriteStream(targetPath));
```
dewen, line 179: previous on events seem to be applied in wrong place, moved to outer function.

beliayeu: @dewen, I believe we need to resolve promise once the file is written to file system
so finish, error events were attached to the stream

packages/bodiless-migration-tool/src/migrationApi.ts
```diff
@@ -62,6 +62,14 @@ type MigrationApiType = {
...
    */
   getSitePath: () => string,
+  /**
+   * return site static directory.
+   */
+  getStaticPath: () => string,
+  /**
+   * Return current page url.
+   */
+  url: string,
```
beliayeu, line 72: I think we already have getPagePath that serves this purpose

packages/bodiless-migration-tool/src/migrationApi.ts
```diff
@@ -62,6 +62,14 @@ type MigrationApiType = {
    * @returns file system directory containing global site data
    */
   getSitePath: () => string,
+  /**
+   * return site static directory.
+   */
+  getStaticPath: () => string,
```
beliayeu, line 68: I think it would be nice to provide ability to pass a resource url and get corresponding fs path

packages/bodiless-migration-tool/src/plugins/imagetags.ts
```diff
@@ -0,0 +1,59 @@
...
+ * creates json file for each image tags
+ * allows to prefix the json file
+ *
+ * @param params to create the image tag scraper plugin
+ */
+const onPageCreate = ({
+  prefix,
+}: ImagetagsFactoryParams) => ({ document, api, downloader }: OnPageCreateParams) => {
+  const pageUrl = api.url;
+  document('img')
```
beliayeu, line 35: do we want to support picture tag?

packages/bodiless-migration-tool/src/plugins/imagetags.ts
```diff
@@ -0,0 +1,59 @@
...
+      const resourceUrl = new URL(asset.attribs.src, pageUrl);
+      // Download with downloaded info
+      const downloaded = await downloader.downloadFiles([resourceUrl.href]);
+      if (downloaded.length >= 1) {
+        const { targetPath } = downloaded[0];
+        const ext = path.extname(targetPath);
+        api.writeJsonFileSync(
+          path.resolve(api.getPagePath(), `${prefix}${path.basename(targetPath, ext)}.json`),
+          {
+            src: targetPath.replace(api.getStaticPath(), ''),
```
beliayeu, line 47: considering encapsulating this logic into api object
DX would be
```
src: api.getResourcePath(resourceUrl);
```

packages/bodiless-migration-tool/tests/downloader.test.ts
```diff
@@ -39,4 +53,23 @@ describe('assets download', () => {
...
+
+  test('Download returns correct path', async () => {
+    const pageUrl = 'https://localhost';
+    const downloadPath = 'static';
+    const downloader = new Downloader(pageUrl, downloadPath);
+    const assets = [
+      'https://localhost/path/to/gatsby.png',
+      'https://external.site.com/images/not-exists.jpg',
+    ];
+    try {
```
beliayeu, line 65: do we need to have try/catch in tests?

beliayeu (changes requested):
1. reconsider downloadFile onEvents
2. include picture tag support
3. consider adjusting migration api

packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -39,58 +39,104 @@ import {
...
           } catch (e) {
             debug(e.message);
           }
+          return {
+            targetPath: '',
+            url: '',
+          };
         },
-        { concurrency: 4 },
+        { concurrency: 1 },
```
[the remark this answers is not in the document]
dewen: that was a debug change, good catch

packages/bodiless-migration-tool/src/downloader.ts
```diff
@@ -39,58 +39,104 @@ import {
...
       );
+      return [...result, ...precheck.EXISTS];
```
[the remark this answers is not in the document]
dewen: i think it depends on what we expect on returns from downloadFiles. I typed it as 

```
type Downloaded = {
  targetPath: string;
  url: string;
};
```

For external url, there will be no saved path, not sure if we need to return, or purpose of that return.

packages/bodiless-migration-tool/src/migrationApi.ts
dewen: @beliayeu I set the get for original page url value (with protocol and host name), might be different from page path.

It's used in packages/bodiless-migration-tool/src/plugins/imagetags.ts

packages/bodiless-migration-tool/tests/downloader.test.ts
dewen: i added the catch to manually fail the test if download promise rejected.

packages/bodiless-migration-tool/src/downloader.ts
dewen: Agreed, onEvents restored

packages/bodiless-migration-tool/src/plugins/imagetags.ts
dewen: per discuss in chat, picture tag is not in scope. Current implementation will process img tag inside picture.

packages/bodiless-migration-tool/src/migrationApi.ts
dewen: It might depend on downloader configuration. I don't see there can be a direct map between resource url and saved path in migration API with actual download the file.

packages/bodiless-migration-tool/src/plugins/imagetags.ts
dewen: added getStaticRelativePath in this case.

beliayeu (dismissed):

packages/bodiless-migration-tool/src/migrationApi.ts
dewen: updated with migrationApi pagePath

beliayeu (dismissed):

wodenx (approved):
All tests passed.

beliayeu (approved):

PR #734, opened 2020-12-03
fix(gatsby-theme-bodiless): allow to pass props to gatsby img

beliayeu:
## Changes
* fixed the issue when image attrs including alt and title are not rendered when gatsby image is activated

### DIFF
--- a/packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
+++ b/packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
@@ -14,9 +14,17 @@
 
 import React, { ComponentType as CT } from 'react';
 import GatsbyImg from 'gatsby-image';
-import type { GatsbyImageProps } from 'gatsby-image';
+import {
+  ifToggledOn,
+  withActivatorWrapper,
+} from '@bodiless/core';
+import type {
+  FluidObject,
+  FixedObject,
+  GatsbyImageOptionalProps,
+} from 'gatsby-image';
 import { Div } from '@bodiless/fclasses';
-import { pick } from 'lodash';
+import { pick, flow } from 'lodash';
 
 type ImageProps = {
   src: string;
@@ -25,16 +33,8 @@ type ImageProps = {
 
 export type GasbyImageProps = ImageProps & {
   preset: string;
-  gatsbyImg?: GatsbyImageProps;
-};
-
-const GatsbyImgWrapper = (props: ImageProps) => {
-  // ToDo: get a list of wrapper activator props in a more maintainable way
-  const rest = pick(props, ['data-bl-activator', 'onClick', 'children']);
-  return (
-    <Div {...rest} />
-  );
-};
+  gatsbyImg?: { fluid: FluidObject | FluidObject[] } | { fixed: FixedObject | FixedObject[] };
+} & GatsbyImageOptionalProps;
 
 // the list is taken from GatsbyImageOptionalProps interface
 const GATSBY_IMG_PROPS = [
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
```diff
@@ -24,24 +33,41 @@ type ImageProps = {
...
+const asGatsbyImage$ = (Component: CT<any>) => {
   const AsGatsbyImage = (props: GasbyImageProps) => {
     const { gatsbyImg, preset, ...rest } = props;
     if (gatsbyImg !== undefined) {
+      const gatsbyImgProps = pick(props, GATSBY_IMG_PROPS) as GatsbyImageOptionalProps;
       return (
-        <GatsbyImgWrapper {...rest}>
-          <GatsbyImg {...gatsbyImg} />
-        </GatsbyImgWrapper>
+        <GatsbyImg {...gatsbyImg} {...gatsbyImgProps} />
```
wodenx, line 70: Do we want these props to override the presets?

packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
```diff
@@ -24,24 +33,41 @@ type ImageProps = {
...
+];
 
-const asGatsbyImage = (Component: CT<any>) => {
+const isGatsbyImage = ({ gatsbyImg }: GasbyImageProps) => gatsbyImg !== undefined;
+
+const asGatsbyImage$ = (Component: CT<any>) => {
   const AsGatsbyImage = (props: GasbyImageProps) => {
     const { gatsbyImg, preset, ...rest } = props;
     if (gatsbyImg !== undefined) {
+      const gatsbyImgProps = pick(props, GATSBY_IMG_PROPS) as GatsbyImageOptionalProps;
```
wodenx, line 68: Do we need to pick these props?  I'm concerned about maintainability here, and the worst case is a site builder will pass a bad prop and get a warning, no?

packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
```diff
@@ -51,4 +77,10 @@ const asGatsbyImage = (Component: CT<any>) => {
   return AsGatsbyImage;
 };
 
+const asGatsbyImage = flow(
+  asGatsbyImage$,
+  ifToggledOn(isGatsbyImage)(withActivatorWrapper('onClick', Div)),
```
wodenx, line 82: what's the use-case for when a gatsby image would not be a gatsby image?

wodenx: Also - we only want this when editable, right?

packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
beliayeu: GatsbyImg component does not accept preset. We need pass image preset variations generated on the backend side. So preset will not have any effect here.

packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
beliayeu: agree. per quick gatsby image code check, it picks props itself, so it looks we are safe here
but will make a quick to validate this

packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
beliayeu: >>what's the use-case for when a gatsby image would not be a gatsby image?
the use case when image is uploaded but gatsby image variations are not generated yet
in this case we are displaying a standard bodiless image

beliayeu: |Also - we only want this when editable, right?

right, will fix it

packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
beliayeu: validated - no warnings are thrown, the invalid props are not rendered in the markup
removed the pick logic in ce9fb9b75a305e327cdf265fd377690f19c9191c

packages/gatsby-theme-bodiless/src/dist/GatsbyImage/asGatsbyImage.tsx
beliayeu: fixed in ce9fb9b75a305e327cdf265fd377690f19c9191c

wodenx (approved):

hvanyo:
@beliayeu  The alt is now being passed but the design isn't.   we can separate out as separate ticket if you wish..

Easiest way to test is is probably switch the test site to having the tout image serve a fluid img and then you will see styling with rounded corners is lost.

hvanyo:
@beliayeu  not sure if its because I am not in a clean env but npm run build is throwing the error: @bodiless/test-site: /Users/heidivanyo/slate/bj-andrei/packages/gatsby-theme-bodiless/dist/fsLogHandler.js:52
@bodiless/test-site:         throw e;

hvanyo (approved):

Issue #645, opened 2020-10-28
Scrape to bodiless data


wodenx:
## User Story
As a site builder
I want to scrape various page-level data from an existing site
So that content entry is quicker

## Acceptance Criteria

AC1. It is possible to configure the migration tool to scrape data from site pages and write it to json files in the directory of the corresponding page, or in the side data directory.

AC2. Configuration of data to scrape may specified either as a list of JS modules or (if configuration files are migrated from .json to.js) as a list of functions.  In the first case, he default export of each modules is a function. In either case, the functionreceives as arguments:
- the parsed HTML for the page
- the page path
- a MigrationApi object which contains methods to write files to the page or site directory.
- an object containing additional parameters specified as part of the configuration.

eg (in migrationconfig.json)
```
{
   ...,
   "dataScrapers": {
      "foo.js",
      {
        "resolve": "bar.js",
        "params": {
          "baz": "bang"
        }
     },
   ...
```
or, in migration-config.js
```
dataScrapers: [
  foo,
  bar({ baz: bang }),
],
```

AC3. The migration tool includesa a predefined scraper for metatags.  This acceps a list of names and a prefix.  For each name in the list it creates a json file at`${prefix}/${name}.json` with the content
```
{
  content: `${value}`
}
```

AC4. There is a reusable migration tool configuration for canvas migration which includes a data scraper (converter) which writes an index.json file using data from the page (eg body class, page-type metatag and/or page path) to set templates correctly for 
- product listing page
- product detail page

scrapeTemplate.js

```
export default ({ document, path, api }) => {
  const pageType = doc.find('meta[name="page-type"]').attr('value');
  const bodyClass = doc.find('body').attr('class').match(/node-type-(.*) /)?[0];
  if (pageType === 'product-listing') {
     api.writeJsonSync('index.json', { '#template': 'product_listing', '#child-template': 'product_detail' });
  } else if (pageType === 'product' || bodyClass === 'product' || pagePath.match(/^\/products)) {
     api.writeJsonSync('index.json', { '#template': 'product_detail' });
  } else if( ... ) {
    ...
  } else {
     api.writeJsonSync('index.json', { '#template': '_default' });
  }
```

  
   ...
```

PR #751, opened 2020-12-10
fix: replace colon in json file names

### LINKED ISSUE #723: Replace colon in json file name
## Description

Some meta file names contain colon character which might cause issues on windows based system. So these colons should be replace with more common characters like hyphens "-".

### Steps to reproduce

- Check json files under folder - ./examples/test-site/src/data/site/

### Expected result

No ":" in json file names

### Actual result

```
./examples/test-site/src/data/site/meta$og:sitename.json
./examples/test-site/src/data/site/meta$og:type.json
./examples/test-site/src/data/site/meta$twitter:card.json
./examples/test-site/src/data/site/meta$twitter:site.json
```

### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

### Affected Version

- If the issue was encountered on a site, please attach your `package.json`
- If the issue was encountered in the Bodiless-JS repo, please link to a
  commit or branch on which the issue occurs.

beliayeu:
## Changes
* replaced colon with hyphen in test-site json files
* updated metatags  nodekeys containing colons
* updated migration tool metatag scraper plugin with colon replacement

## Related Issues

Fixes #723


VancheeZze (approved):

packages/bodiless-migration-tool/src/plugins/metatags.ts
```diff
@@ -47,10 +47,15 @@ const onPageCreate = ({
     .toArray()
     .filter(item => item.attribs.name !== undefined && item.attribs.content !== undefined)
     .filter(item => !names || names.includes(item.attribs.name))
+    .map(item => ({
+      // replace colon with hyphen to avoid issues on windows
```
dewen, line 51: not sure if we need loop through all items for replacement, instead maybe just replace inline on LINE56:
```
const filename = item.attribs.name || item.attribs.property;
...
path.resolve(api.getPagePath(), `${prefix}$${filename.replace(':', '-')}.json`),
```

this way we could handle additional tag attribute process.

beliayeu: hey @dewen, nice catch
I think it will lead also to changing the filtering logic (defined in line 48)
currently, the metatag plugin supports only standard attributes, it supports only metatags that have 'name' and 'content'. ('property' is not a standard attribute, see https://html.spec.whatwg.org/multipage/semantics.html#the-meta-element)
per discussion in #645, the decision was taken to implement support of non-standard attributes including 'property' in a separate ticket. I guess you created a ticket for it.
lets discuss if we want doing it as part of this ticket

dewen (approved):

PR #764, opened 2020-12-16
doc(core): Update node version requirement documentation.

wodenx:
## Changes

- Clearly indicate that only Node 12.x is supported.


packages/bodiless-documentation/doc/README.md
```diff
@@ -29,9 +29,7 @@ npm run start
 
 Then Visit `localhost:8005` in your browser.
 
-> Requires:
-> - [NodeJS](https://nodejs.org/en/download/) (v12.18.0+ preferred)
-> - [NPM](https://docs.npmjs.com/) version 6.13.1+
+> Requires [NodeJS](https://nodejs.org/en/download/) (v12.x only)
```
dewen, line 32: v12.19.x ?

wodenx: i think ppl can use an older 12.x version if they need to

wodenx: not sure why they would

dewen: right, we don't specify engine-strict

dewen (approved):

hvanyo (commented):
Its also mentioned here examples/starter/README.md

hvanyo (approved):

PR #769, opened 2020-12-17
feat(richtext): paste links into richtext

### LINKED ISSUE #646: Preserve links when pasting into Rich Text editor
## User Story
As a site builder
I want to preserve links when I paste HTML into a rich text editor
So I don't have to enter them manually

## Acceptance Criteria
AC1. When a user pastes HTML containing links into a rich text editor, the links are preserved, including their href data.
AC2. For a given rich text editor, the site builder can specify what component should be used to render the link.
AC2a. On the test site add a test case to the RTE page so that the following can be verified 
When a user uses the RTE to paste in content the link component is used for rendering the href data AND the user can use another RTE to see the differences 
AC3. When the link data is saved they are normalized using what is defined for the link component that was specified in 

## Notes
This should be treated as the first step in a more general HTML deserializer for rich text editors.
Scope: When a user pastes content from one RTE to another RTE, changes in this behavior is out of scope. If there are any issues/notes create a gap ticket for it.

beliayeu:
## Changes
* added html deserializer to richtext so that when a user pastes HTML containing links into a rich text editor, the links are preserved, including their href data.
* added support for Bold, Italic, Strikethrough, Header1, Header2, Header3 html elements
* added a sample migration tool plugin that demonstrates how html deserializers can be leveraged in the migration case

## Related Issues

Closes #646


packages/bodiless-richtext/src/serializers/defaultDeserializers.ts
```diff
@@ -0,0 +1,97 @@
...
+  Bold: createBoldDeserializer(),
+  Italic: createItalicDeserializer(),
+  Link: createLinkDeserializer(),
+  StrikeThrough: createStrikeDeserializer(),
+  Header1: createHeader1Deserializer(),
+  Header2: createHeader2Deserializer(),
+  Header3: createHeader3Deserializer(),
+});
+
+export {
```
dewen, line 88: Shall we move export to deserializer definition? Saves a few lines

dewen (approved):

dewen (approved):

VancheeZze (approved):

PR #775, opened 2020-12-21
feat(layout): fix rtl flow container. 

awm086:
## Changes:
- Fix flow container on rtl pages.
- Allow for a rtl flow container.
## Test Instructions:
- Visit /flow-container-rtl
- add a component 
- note the drag handle on the left side working as expected.

## Related Issue:
N/A


packages/bodiless-ui/src/elements.tsx
```diff
@@ -180,6 +180,13 @@ export const ResizeHandle = flow(
   withChild(() => <ResizeHandleBar />),
 )(Div);
 
+export const ResizeHandleRTL = flow(
+  addClasses(
+    'bl-flex bl-bg-red-600 bl-justify-center bl-w-5 bl-left-rem-1 bl-z-1 bl-h-three-quarters bl-relative bl-top-half bl-transform bl--translate-y-1/2',
+  ),
+  withChild(() => <ResizeHandleBar />),
+)(Div);
+
```
VancheeZze, line 189: I think it is better to re-use `ResizeHandle`:

```js
export const ResizeHandleRTL = flow(
  addClasses('bl-left-rem-1'),
  removeClasses('bl-right-rem-1'),
)(ResizeHandle);
```

Also, I'm not sure if we need `bl-bg-red-600` for the handle, was that per design?

VancheeZze (changes requested):

packages/bodiless-ui/src/elements.tsx
awm086: good catch.

VancheeZze (approved):

TatsianaMiraniuk (approved):

Issue #417, opened 2020-07-02
Image component: Unhandled Rejection (TypeError) is thrown when dragging non-image file or more than one image


hvanyo:
## Description

Describe the issue that you're seeing.

### Steps to reproduce

1. Open a page with Image component, e.g. /touts page
2. Click on Image placeholder to get a tooltip bar
3. Click on Image icon to open Image Uploading form
4. In your machine, select two images and drag them into 'Drag a file or click here to upload.' area

### Expected result

the case should be handled correctly without crashing and the end-user should get an appropriate message in Image uploading form

### Actual result

'Unhandled Rejection (TypeError): Cannot read property 'name' of undefined' error is thrown.

5. Refresh the page
6. Click on Image placeholder to get a tooltip bar
7. Click on Image icon to open Image Uploading form
8. In your machine, select non-image file, e.g. *.pdf and drag it into 'Drag a file or click here to upload.' are

### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

### Affected Version

- If the issue was encountered on a site, please attach your `package.json`
- If the issue was encountered in the Bodiless-JS repo, please link to a
  commit or branch on which the issue occurs.

PR #776, opened 2020-12-21
fix(components): Image component: Fix Unhandled Rejection

VancheeZze:
## Changes
- Fix Unhandled Rejection error when uploading multiple images or non-image type.

## Test Instructions
 - See STR here: https://github.com/johnsonandjohnson/Bodiless-JS/issues/417

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/417


packages/bodiless-components/src/Image.tsx
```diff
@@ -104,6 +104,13 @@ export function DropZonePlugin({ formApi, targetFieldName, ui }: {
   });
 
   const onDrop = useCallback(acceptedFiles => {
+    // When files are rejected by the react-dropzone,
+    // acceptedFiles is an empty array.
+    if (acceptedFiles.length < 1) {
+      setStatusText('File type not accepted or too many, try again!');
```
awm086, line 110: maybe `the file type you are trying to upload is not supported?` and do we know when it's too many perhaps we give a different message?

VancheeZze: This is the default message from line 69 ( https://github.com/johnsonandjohnson/Bodiless-JS/pull/776/files#diff-e254e55e5e7a340d948c110dec56bf42894f6659299478be4c6b8b0195ce4f85R69 ). I think it is ok to keep it as is

awm086 (approved):

TatsianaMiraniuk (approved):

PR #784, opened 2021-01-04
feat(components): Allow sanitizers for editable text.

### LINKED ISSUE #796: Allow Sanitizers for Editable Text
## User Story
As a site builder
I want to be able to add sanitizers/transformers to plain-text editors
So that I can support simple special formatting

## Acceptance Criteria
AC1. It is possible to configure a plain text editor to add simple formatting to the entered text.
AC2. The formatting will only apply when the editor is not focused in edit mode.

## Notes
- Primary use-case is auto-superscripting.  This is demonstrated in the site footer copyright on the test site.

wodenx:
## Changes
- Enhance asEditable to allow specifying a "sanitizer" function which
  will be invoked on render to process the text.
- Demonstrate use of this API by auto-superscripting copyright symbols
  in plain-text editors on the test-site.

## Related Issues
Closes #796


examples/test-site/src/components/Elements.token.ts
```diff
@@ -85,7 +86,17 @@ const asImageRounded = addClasses('rounded-lg');
...
 /* Edit component */
-const asEditable = asEditableCore;
+const asEditable = (nodeKeys?: WithNodeKeyProps, placeholder?: string) => asEditableCore(
+  nodeKeys,
+  placeholder,
+  // Overrides to add auto-superscript.
+  () => ({
+    sanitizer: (html: string) => html
+      .split('')
+      .map(c => ('©'.includes(c) ? `<sup>${c}</sup>` : c))
```
beliayeu, line 96: per my understanding if the character is already superscripted then it will superscripted twice

examples/test-site/src/components/Elements.token.ts
```diff
@@ -85,7 +86,17 @@ const asImageRounded = addClasses('rounded-lg');
 const asEditableLink = asBodilessLink;
 
 /* Edit component */
-const asEditable = asEditableCore;
+const asEditable = (nodeKeys?: WithNodeKeyProps, placeholder?: string) => asEditableCore(
+  nodeKeys,
+  placeholder,
+  // Overrides to add auto-superscript.
+  () => ({
```
beliayeu, line 93: what do you think about passing "sanitizer" as a prop instead of using useOverrides pattern?

packages/bodiless-components/src/Editable.tsx
```diff
@@ -26,50 +28,68 @@ import {
...
+const EditableText = observer((props: EditableProps) => {
+  const { node } = useNode<EditableData>();
+  const { placeholder = '', useOverrides = () => ({}) }: EditableProps = props;
+  const { sanitizer = identity }: EditableOverrides = useOverrides(props);
+  const text = (node.data.text !== undefined ? node.data.text : props.children) || '';
+  const [hasFocus, setFocus] = useState(false);
+  const ref = useRef<HTMLElement>(null);
+  const onChange = () => {
+    const newText = ref.current?.innerHTML || '';
+    const plainText = newText.replace(/(<([^>]+)>)/gi, '');
```
beliayeu, line 66: concerning that we might strip some valid text
'some text <20%. other text >10%' -> 'some text 10%'

packages/bodiless-components/src/Editable.tsx
```diff
@@ -26,50 +28,68 @@ import {
...
+    node.setData({ text: plainText });
   };
+  const onFocus = () => { setFocus(true); };
```
beliayeu, line 69: consider wrapping with useCallback

packages/bodiless-components/__tests__/Editable.test.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { flow } from 'lodash';
+
+import { render } from 'enzyme';
+import { withMockNode } from './helpers/MockContentNode';
+import { asEditable } from '../src';
+
+describe('asEditable', () => {
+  describe('When not editable', () => {
```
beliayeu, line 9: it would be great to provide a test case in which multiple sanitizers are added

examples/test-site/src/components/Elements.token.ts
```diff
@@ -85,7 +86,17 @@ const asImageRounded = addClasses('rounded-lg');
...
+    sanitizer: (html: string) => html
```
beliayeu, line 94: do we want to handle use case and provide DX re how to add a new sanitizer without losing existing sanitizer?

examples/test-site/src/components/Elements.token.ts
wodenx: that is correct, but this is a plain text editor, so it should not have html in the data.

packages/bodiless-components/src/Editable.tsx
wodenx: Well - we are getting the inner html and then removing any tags.  Characters you type are already escaped and won't be removed.  However, this is probably unnecessary, as we are not running the sanitizers when in edit mode, so there should be no html to strip.  Removing it.

examples/test-site/src/components/Elements.token.ts
wodenx: per discussion, deferring this to a later iteration.

packages/bodiless-components/src/Editable.tsx
wodenx: done

packages/bodiless-components/__tests__/Editable.test.tsx
wodenx: not sure there is anything specific to sanitizers here - see below.

examples/test-site/src/components/Elements.token.ts
wodenx: I think the pattern would be:
```
const asSuperScriptEditable = (nodeKey, placeholder, useOverrides) => {
  const useOverrides$ = (props) => {
    const overrides = useOverrides(props);
    const { sanitizer: nextSanitizer } = overrides;
    const newSanitizer = (text) => {
      const t = addSuperscript(text);
      return nextSanitizer(t);
    };
    return { ...overrides, sanitizer };
  }
  return asEditable(nodeKey, placeholder, useOverrides$);
}
```
There's nothing specific here to this - though I agree the pattern should be documented

beliayeu (approved):

examples/test-site/src/components/Elements.token.ts
hvanyo, line 96: Can you add ® as test case as well as that is more likely to be used

hvanyo (approved):

PR #792, opened 2021-01-09
fix(richtext): #791 - Fix crash when pasting empty last element.

### LINKED ISSUE #791: Pasting empty paragraph in RTE causes crash
STR
Copy HTML with an empty last element to clipboard.  Easiest way to do this on a mac is to triple click to select a line of html - osx will insert an empty <br> tag at the end of the markup.

ER
Text pastes properly

AR
Page crashes with error: "Cannot get the start point in the node at path [x] because it has no start text node."

Notes:
See: https://github.com/ianstormtaylor/slate/issues/3625

wodenx:
## Changes
- Ensure richtext deserializer does not product elements with empty children.

## Related Issues
Fixes #791.


beliayeu (approved):

hvanyo:
https://edit.5znkk2cxg4-jvaff4bu65vgm.us-2.platformsh.site/

hvanyo (approved):

PR #832, opened 2021-02-01
feat(search): Search Component - Type Ahead

beliayeu:
## Changes
* List of potential keyword phrases appears with the first letter typed as a list below the search box (see screenshot attached) 
* Order is based on frequency and lists no more then 5 items
* Items list # of results 
* Item list is at full width of the search field
* search results do not contain punctuation 
* Responsive behavior should be the same on all formats


examples/test-site/src/components/Search/token.tsx
```diff
@@ -48,12 +64,34 @@ const responsiveSearchDesign = {
     withIcon('search'),
     addClasses('flex absolute right-0 self-center mr-4'),
   ),
+  Suggestions: withDesign({
```
dewen, line 67: Suggestions with design seems duplicated with Line 45-55, suggest merge into a function.

packages/bodiless-search/src/SearchClient.ts
```diff
@@ -46,6 +59,27 @@ class SearchClient implements SearchClientInterface {
     return this.searchEngine.search(filtered);
   };
 
+  suggest = (queryString: string, settings: SuggestionSettings = defaultSuggestSettings) => {
```
dewen, line 62: suggest to define the return type as lunr.TokenSet[]

packages/bodiless-search/src/components/Search.tsx
```diff
@@ -215,6 +229,11 @@ const SearchBoxBase: FC<SearchProps> = ({ components, ...props }) => {
         placeholder={placeholder}
       />
       <SearchButton onClick={onClickHandler} />
+      {
+        queryString !== ''
+        // @ts-ignore
```
dewen, line 234: what type msg is ignored here?

dewen (dismissed):

examples/test-site/src/components/Search/token.tsx
beliayeu: makes sense, merged in 37be9adaf346eb4c100f2a3342e74dd0c57388e5

packages/bodiless-search/src/SearchClient.ts
beliayeu: return type here is array of [Suggestion](https://github.com/johnsonandjohnson/Bodiless-JS/blob/37be9adaf346eb4c100f2a3342e74dd0c57388e5/packages/bodiless-search/src/types.ts#L71) that contains a collection of terms and its frequency. so I think the type is fine here

packages/bodiless-search/src/components/Search.tsx
beliayeu: enhanced types in 37be9adaf346eb4c100f2a3342e74dd0c57388e5

dewen (dismissed):

dewen (dismissed):

dewen (dismissed):

examples/test-site/src/components/Search/token.tsx
```diff
@@ -36,10 +36,10 @@ const isEven = (item: number) => item % 2 === 0;
 const isOdd = (item: number) => item % 2 === 1;
 
 const withSuggestionsDefaultDesign = withDesign({
-  Wrapper: addClasses('absolute top-full z-50 w-full'),
+  Wrapper: addClasses('absolute top-full z-50 w-full border border-black'),
```
hvanyo, line 39: Can you just do this at top of this Search/token.ts file:
asSuggestionsborder = addClasses(...) and then add 
same for hovers, and focus

beliayeu: fixed in 797013394b7a6bf2f7a7b276ca548ebc86509f1b
@hvanyo, thanks

dewen (approved):

konst8 (approved):

Issue #797, opened 2021-01-12
Make text editors designable


wodenx:
## User story
As a site builder
I want to be able to customize text editors which have been added to compound components
So I can more reasily reuse them.

## Acceptance Criteria
AC1. The existing `withChild` hoc is refactored or replaced so that it is possible to alter the added child using the design API.
AC2. The existingHOC's at site and package level which add plain pr rich text editors are refactored so to use the above withChild, so that it is possible to customize them.

## Example
```
const withDefaultToutEditors = withDesign({
 Title: asEditable('title', 'Title'),
  Body: withEditorBasic('body', Body'),
   ...
}

const withSpecialToutEditors = flow(
  withDefaultToutEditors,
  withDesign({
    Title: withDesign({
      Editor: replaceWith(EditorSimple),  // Make it a rich text editor instead of a plain text
    },
    Body: withDesign({
      Editor: withDesign({
        H2: asHeader2,  // Add another formatting option.
      }),
    }),
  }),
);
```

PR #837, opened 2021-02-04
feat(core): designable child API

beliayeu:
## Changes
* replaced existing withChild hoc so that it is possible to alter the added child using the design API.

## Test Instructions
* for testing purposes, a new richtext called "Alphabetic Full Rich Text"  is added to /richtext page of test-site. Leveraging the implemented API, the richtext accepts only alphabetic characters, space, backspace, delete and enter characters. Please note that the richtext configured so that the logic to disallow characters works only during typing (it still will be possible to insert non-alphabetic characters using pasting)

## Related Issues

Fixes: #797

### DIFF
--- a/examples/test-site/src/data/pages/richtext/index.tsx
+++ b/examples/test-site/src/data/pages/richtext/index.tsx
@@ -14,14 +14,39 @@
 
 import React from 'react';
 import { graphql } from 'gatsby';
+import flow from 'lodash/flow';
 import {
   NodeViewer,
 } from '@bodiless/components';
+import { withDesign, addProps, Div } from '@bodiless/fclasses';
 import { Page } from '@bodiless/gatsby-theme-bodiless';
 
 import Layout from '../../../components/Layout';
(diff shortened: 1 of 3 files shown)


examples/test-site/src/data/pages/richtext/index.tsx
```diff
@@ -14,14 +14,39 @@
...
+} from '../../../components/Editors';
+
+const AlphabeticFullFeaturedEditor = flow(
+  withEditorFullFeatured('alphabeticRTE', 'Type something here...'),
+  withDesign({
+    Child: addProps({
+      onKeyDown: event => {
+        if (
+          // alphabet characters
+          !(event.which >= 65 && event.which <= 120)
```
SW999, line 40: Not critical, but I'd recommend not to use 'which'.
Please look at https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which

SW999 (approved):

VancheeZze (approved):

PR #840, opened 2021-02-08
chore: update packages dependencies

SW999:
## Changes

* Updated @types/node to ^12
* Fixed vulnerabilities in npm modules
* Tailwind upgraded to ^2.0.2
* Removed *future* options from tailwind.config.js as not supported
* Updated renamed utility classes (*-no-wrap => *-nowrap, .clearfix => .flow-root)
* Added *screens* to tailwind.config.js to prevent unexpected breacpoint on *.container* class
* Updated *colors* in tailwind.config.js for default colors to prevent braking changes in default colors
* Added *fontSize* in tailwind.config.js for default fon-sizes to prevent adding unexpected line-heights
* Updated *borderRadius* option


examples/starter/package.json
```diff
@@ -29,6 +29,7 @@
     "@bodiless/richtext-ui": "^0.0.68",
     "@bodiless/search": "^0.0.68",
     "@bodiless/ui": "^0.0.68",
+    "autoprefixer": "^10.2.4",
```
VancheeZze, line 32: I also think we should move `autoprefixer` and `postcss` into `peerDependencies` as per Tailwind upgrade guide

SW999: I tried to move these dependencies into peerDependencies. In this scenario, the application does not start.

SW999:
Outdated

Issue #616, opened 2020-10-22
Highlighting Link in Tout replaced it with actual href link ON Starter Site


hvanyo:
Repeat of : https://github.com/johnsonandjohnson/Bodiless-JS/issues/161
Fix: https://github.com/johnsonandjohnson/Bodiless-JS/pull/306 on test-site
and needs to be addressed on starter site as its replicating through site builds

**NOTE**:  DURING discussion, if we continue on this pattern we ahve to add this workaround/fix at everypoint a link is used.  It was determined to move this fix to the link package level so it takes effect for all link cases.

## Description

If you try highlight a linked tout cta, it replaces it with the actual href link

### Steps to reproduce

1. place a tout and link the cta and name the cta link (or use existing tout that has linked CTA)
2. highlight the CTA link (to copy it elsewhere)

### Expected result

Highlights as expected

### Actual result

Highlighting injects the href value into the text cta title.
https://www.screencast.com/t/qbmrsFj4f8Su

### Environment

Run `gatsby info --clipboard` in your project directory and paste the output here.

asiarheyeu:
closed as completed

PR #844, opened 2021-02-10
fix(starter): Highlighting Link in Tout replaced it with actual href link

SW999:
## Changes
* fixed the bug #616 by making cta link non-draggable when in edit mode


hvanyo (commented):
@SW999 
I think we will want this fix into into Link package -> https://github.com/johnsonandjohnson/Bodiless-JS/blob/master/packages/bodiless-components/src/Link/asBodilessLink.tsx#L138 

rather than all these site specific errors.  Fix once at package level and then we don't have to repeat this code everywhere.   While this bug creates in Tout link, it could happen wherever a link is used.

packages/bodiless-components/src/Link/asBodilessLink.tsx
```diff
@@ -138,6 +139,9 @@ const asBodilessLink: AsBodilessLink = (
   ifEditable(
```
konst8, line 139: @SW999 Can we avoid using ifEditable twice here?

SW999: Fixed

konst8 (approved):

TatsianaMiraniuk (approved):

PR #845, opened 2021-02-10
chore: update packages dependencies

SW999:
## Changes

* Updated @types/node to ^12
* Fixed vulnerabilities in npm modules
* Tailwind upgraded to ^2.0.2
* Removed *future* options from tailwind.config.js as not supported
* Updated renamed utility classes (*-no-wrap => *-nowrap, .clearfix => .flow-root)
* Added *screens* to tailwind.config.js to prevent unexpected breacpoint on *.container* class
* Updated *colors* in tailwind.config.js for default colors to prevent braking changes in default colors
* Added *fontSize* in tailwind.config.js for default fon-sizes to prevent adding unexpected line-heights
* Updated *borderRadius* option


examples/starter/tailwind.config.js
```diff
@@ -43,7 +39,124 @@ module.exports = mergeWithBodilessConfigs({
     |
     */
 
-    // colors: {},
+    colors: {
```
hvanyo, line 42: @SW999  I would prefer to only pull colors over that we are using in starter kit

examples/test-site/tailwind.config.js
```diff
@@ -43,7 +39,124 @@ module.exports = mergeWithBodilessConfigs({
     |
     */
 
-    // colors: {},
+    colors: {
```
hvanyo, line 42: This is fine declaring them all here as this site is start of new site.

packages/bodiless-ui/bodiless.tailwind.config.js
```diff
@@ -205,16 +205,97 @@ module.exports = {
...
         'grey-600': '#718096',
+        'grey-700': '#4a5568',
         'grey-800': '#2d3748',
         'grey-900': '#1a202c',
 
         red: '#e3342f',
         green: '#309795',
 
         'black-transparent': '#00000066',
+
```
hvanyo, line 221: I am not sure its necessary to add the colors.  Our admin palette should only be using specific colors:  i.e. https://jhdfdc.axshare.com/#g=1&p=admin_pallette&id=eqb1xp 

I would say we do not have to add the ones we are not using and keep this cleaner.   I believe they are purged if not in use but its more so others don't use these colors in the admin.   



@VancheeZze  do you agree?

VancheeZze: I 100% agree, just wanted to add that the classes defined here are all prefixed with `bl-` so there are really just a few colors that we need to keep here. The other color classes like `bg-blue` would be taken from the site's tailwind config.

packages/bodiless-ui/bodiless.tailwind.config.js
```diff
@@ -205,16 +205,97 @@ module.exports = {
...
         'grey-100': '#f7fafc',
         'grey-200': '#edf2f7',
+        'grey-300': '#e2e8f0',
         'grey-400': '#cbd5e0',
+        'grey-500': '#a0aec0',
         grey: '#a0aec0',
         'grey-600': '#718096',
+        'grey-700': '#4a5568',
         'grey-800': '#2d3748',
         'grey-900': '#1a202c',
```
VancheeZze, line 215: And I would double-check what `bl-` prefixed gray classes we really use and get rid of all others, format it the same way as colors below and if possible rename from `grey` to `gray`:

```
gray: {
  100: '#f7fafc',
  300: '#e2e8f0',
  ...
}
```

packages/bodiless-ui/bodiless.tailwind.config.js
hvanyo: We don't have to fix md files but we may have to change those colors in test to something we are keeping in tailwind.  Pretty sure color is irrevalent there.  :)

examples/starter/tailwind.config.js
SW999: Fixed

packages/bodiless-ui/bodiless.tailwind.config.js
SW999: Fixed. I found we use some colours not relevant to https://jhdfdc.axshare.com/#g=1&p=admin_pallette&id=eqb1xp. They were added more than a year ago. For example, we use **primary: '#0070c8'**, but in the palette we have **#3182CE**

packages/bodiless-ui/bodiless.tailwind.config.js
SW999: Done

VancheeZze (dismissed):

hvanyo (approved):

VancheeZze (dismissed):

examples/test-site/package.json
```diff
@@ -39,7 +39,7 @@
     "gatsby-plugin-canonical-urls": "^2.1.18",
     "gatsby-plugin-google-tagmanager": "^2.1.4",
     "gatsby-plugin-sitemap": "^2.2.24",
-    "informed": "^3.9.0",
+    "informed": "~3.27.0",
```
VancheeZze, line 42: Why did we have to downgrade informed?

VancheeZze (approved):

TatsianaMiraniuk (approved):

PR #861, opened 2021-02-17
feat(bodiless-core-ui): activate context for group in Local context menu

SW999:
## Changes
* added event listener on local menu group title to activate item on click
* underlined title of leftmost group
* added style for underline not active elements on hover

## Test Instructions
* open Homepage 
* select Edit mode
* click on Image inside block "Giving back to community"
* make sure the conditions are met:
AC1. When you click the group title on the local context menu, the corresponding context is activated
AC2. There is a hover effect on the group title which indicates an interaction (underline).
AC3. The left most (inner most) group title is always underlined.

Group title will be **not clickable** and **without any additional styles** in case a group contains only one element.

### DIFF
--- a/packages/bodiless-core-ui/src/LocalContextMenu.tsx
+++ b/packages/bodiless-core-ui/src/LocalContextMenu.tsx
@@ -15,7 +15,12 @@
 import React, { FC } from 'react';
 import ReactTooltip from 'rc-tooltip';
 import { flow } from 'lodash';
-import { addClasses, addProps, removeClasses } from '@bodiless/fclasses';
+import {
+  addClasses,
+  addClassesIf,
+  addProps,
+  removeClasses,
+} from '@bodiless/fclasses';
 import {
   ContextMenu, ContextMenuProps, ContextMenuUI, IContextMenuItemProps,
 } from '@bodiless/core';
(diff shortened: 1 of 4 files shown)


packages/bodiless-core-ui/src/LocalContextMenu.tsx
```diff
@@ -51,19 +56,27 @@ const LocalTooltip: FC<ReactTooltip['props']> = props => (
...
+}) => {
   const hidden: boolean = Boolean(option && (
     typeof option.isHidden === 'function' ? option.isHidden() : option.isHidden
   ));
   if (hidden) return null;
+  const { context, label } = option || {};
+  const onClick = context ? { onClick: () => context.activate() } : {};
+  const GroupTitle = flow(
+    removeClasses('bl-mb-grid-2 bl-min-w-xl-grid-1'),
+    addClassesIf(() => Number(index) > 0)('hover:bl-underline cursor-pointer'),
```
VancheeZze, line 72: Lets change `cursor-pointer` to `bl-cursor-pointer` as well

SW999: Updated

VancheeZze (approved):

konst8 (approved):

VancheeZze (approved):

PR #912, opened 2021-03-09
Fix Jamstack app crashes if git command return error

dewen:


examples/test-site/src/data/pages/bulletpoints$default$bullet.json
```diff
@@ -1,3 +1,3 @@
 {
```
beliayeu, line 1: I believe we can remove this file change from the commit so that keep git history clean

packages/bodiless-backend/src/git.js
```diff
@@ -246,21 +250,32 @@ const getConflicts = async (target = 'upstream') => {
...
-    );
-    await GitCmd.cmd()
-      .add('add', '--all')
-      .exec();
-    await GitCmd.cmd()
-      .add('commit', '-m', 'TEMPORARY COMMIT')
-      .exec();
+
+    try {
+      const result = await new Promise((resolve, reject) => {
```
beliayeu, line 255: tentatively
I think we can use [util.promisify](https://nodejs.org/docs/latest-v12.x/api/util.html#util_util_promisify_original)

packages/bodiless-backend/src/git.js
```diff
@@ -121,33 +121,37 @@ const uncommitted = async () => {
...
+      production: { branch: productionBranch, ...result[1] },
+      local: {
+        branch,
+        commits: [...result[2].commits],
+        files: [...result[2].files, ...result[3].files],
+      },
     };
+    return status;
+  } catch (e) {
+    throw new Error(`Error occurred: ${e.message}`);
```
beliayeu, line 153: if an error happens, will we catch it? or will it lead to the backend is down

packages/bodiless-backend/src/GitCmd.js
```diff
@@ -61,6 +61,9 @@ class GitCmd {
       cmd.stderr.on('data', data => {
         stderr += data.toString();
       });
+      cmd.on('error', (err) => {
```
beliayeu, line 64: am I right that we need `cmd.stderr.on` and `cmd.on('error'` both

dewen: My understanding is ```.on('error')``` is used to add handler for failed spawn cmd, while ```.stderr.on('data', ...)``` for command err output.

So we need both error types handled here

packages/bodiless-backend/src/git.js
dewen: this is caught from upper level, in backend.js route exception handler.

packages/bodiless-backend/src/git.js
dewen: Updated, thx

beliayeu (approved):

TatsianaMiraniuk (approved):

PR #916, opened 2021-03-15
feat(layouts): Supply HTML to Flow Container

### LINKED ISSUE #828: Supply HTML to Flow Container
## User Story
As a content editor
I want to be able to paste HTML to a flow container
So that I can work more quickly.

## Use Case
- Quickly pasting HTML containnig unsupported structures (eg lists) to a flow container
- Processing data from an exernal source into a flow container

## Acceptance Criteria

AC1. There is an HTML deserialization utility available for flow containers.  This accepts a list of deserializers along with the design key specifying a component which should be used to render each.

AC2. The deserializer can be provided to any flow container by wrapping the flow container in an HOC.   This HOC accepts html as a prop and produces default content for the flow container.  The default content consists of a set of items corresponding to the discrete blocks processed by the hoc. For each item:
- The width is set to full
- The type is set to the component type specified for the deserializer which applied to that block
- The child node for the item is provided with the output of the deserializer 

AC3. There is a page on the test site, that has an input field that populates the flow container to test the above functionality. 

AC4. Deserializers are provided for RTE and List components

beliayeu:
## Changes
* added ability to supply html to flow container

## Test Instructions
* open /flow-container/html-deserializer/ page on test-site

## Related Issues

Closes #828


packages/bodiless-layouts/src/deserializers/createFlowContainerItem.ts
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-disable import/prefer-default-export */
+
+import { v4 } from 'uuid';
+
+export const createFlowContainerItem = (type: string) => ({
```
dewen, line 19: Export default?

packages/bodiless-core/src/Contentful/withDefaultContent.tsx
```diff
@@ -34,17 +34,19 @@ import { DefaultContentNode } from '../ContentNode';
  * @returns
  * An HOC providing default content to the wrapped component.
  */
-const withDefaultContent = <P extends object, D extends object>(content: D|((props:P) => D)) => (
+// eslint-disable-next-line max-len
+const withDefaultContent = <P extends { content?: D}, D extends object>(content: D|((props:P) => D)) => (
```
dewen, line 38: ```{ content?: D},``` add space after D,
```{ content?: D },```

File: packages/bodiless-layouts/src/deserializers/createFlowContainerItem.ts (hunk 1 above)
Comment (dewen) [reply packages/bodiless-layouts/src/deserializers/createFlowContainerItem.ts:19]: also can we create a type for createFlowContainerItem?

Comment (dewen) [reply packages/bodiless-layouts/src/deserializers/createFlowContainerItem.ts:19]: should be ```FlowContainerItem```?

File: packages/bodiless-layouts/src/deserializers/listDeserializer.ts [hunk 3]
```diff
@@ -0,0 +1,73 @@
...
+            text: listItem.textContent,
+          },
+        };
+      }
+    }
+  });
+  return result;
+};
+
+const createListDeserializer = (type: string) => ({
```
dewen, line 63: If type is always ```List```, we might not need to pass type param here.

packages/bodiless-layouts/src/deserializers/createFlowContainerItem.ts
beliayeu: fixed in 226a3db12ba27d41b336b988e0b629427cdb120c

packages/bodiless-core/src/Contentful/withDefaultContent.tsx
beliayeu: fixed in 226a3db12ba27d41b336b988e0b629427cdb120c

packages/bodiless-layouts/src/deserializers/listDeserializer.ts
beliayeu: I think there can be different types of lists like ChameleonList, BasicList, etc
the idea to have this function is to provide a match function responsible for detecting lists in html

dewen (dismissed):

hvanyo (approved):

dewen (approved):

dewen (approved):

PR #973, opened 2021-04-24
Fix wsod when adding flow container item.

wodenx:
## Changes
- Add a check to be sure value is not undefined when migrating slate data.

### STR
1. Add an RTE to flow container
2. Type some text, highlight, make a link
3. Click on the link, then click hte Add (+) button to add a component to the fc.
4. 
### ER
Component selector opens normally

### AR
WSOD.

```
TypeError: Cannot read property 'document' of undefined
(anonymous function)
src/withDataMigrator.tsx:94
  91 | const withDataMigrator = (Component: any) => (
  92 |   ({ value, ...rest }: any) => {
  93 |     let newValue = value;
> 94 |     if (value.document !== undefined) {
  95 |       const nodes = removeLeaves(value.document.nodes);
  96 |       newValue = nodes.map(migrateNode) || [];
```

### DIFF
--- a/packages/bodiless-richtext/src/withDataMigrator.tsx
+++ b/packages/bodiless-richtext/src/withDataMigrator.tsx
@@ -91,7 +91,7 @@ const migrateNode = (oldNode: Node) => {
 const withDataMigrator = (Component: any) => (
   ({ value, ...rest }: any) => {
     let newValue = value;
-    if (value.document !== undefined) {
+    if (value?.document !== undefined) {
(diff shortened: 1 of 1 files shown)


wodenx:
I think this is related to the fact that we are passing the current node to all items in the component selector preview, which can have unpredictable results if the data in the node does not match what hte compomnents expect.  It's probably a good idea to ensure that components can deal with bad data gracefully - or to put an error boudary around all flow container items.

PR #1015, opened 2021-06-21
feat(navigation): Menu Accessibility

VancheeZze:

### DIFF
--- a/packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
+++ b/packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
@@ -13,8 +13,7 @@
  */
 
 import React, { ComponentType, FC } from 'react';
-import { flow } from 'lodash';
-import { withPrependChild, ifToggledOn, useNode } from '@bodiless/core';
+import { withParent, withPrependChild, useNode } from '@bodiless/core';
 import type { LinkData } from '@bodiless/components';
 import {
   addProps,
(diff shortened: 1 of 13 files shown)


packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * HOC that wrappes the component in a navigation region implemented with
+ * a nav element that has an aria-label that matches the label on the menubar.
+ */
+const withMenuNav = asToken(
+  withParent(Nav, 'Nav'),
+  withDesign({
+    Nav: asToken(
+      addClasses('w-full'),
+      addProps({ 'aria-label': 'Main Site Navigation Menu' }),
```
marcopagliarulo, line 119: I am not sure is proper to set aria-label in that way. At site level, if I add more than one accessible menu,  everyone would have the same aria-label.

hvanyo:
```suggestion
      addProps({ 'aria-label': 'Menu' }),
```

hvanyo: @marcopagliarulo  so this doesn't stagnant and we keep this moving while its out, would you consider we just change the arial label to 'Menu' so its generic and applies to all menus.   If we find it needs to be set per menu we come back and gap it.

marcopagliarulo: In order to move forward I think we can even keep the current aria label. Any label would be potentially wrong, because the menu has navigation role, that is inherited by the Nav tag. It means that with two or more menu we would have many equal landmark, which should be unique. When the same tag with a default landmark is on the page, the landmark is defined by the aria label.

I think the best solution would be to have an hidden element, and then use aria-labelledby to reference this element, then the element would be editable, in order to allow setting the aria label on edit interface.

If you agree, I don't see any issue to handle it as a gap

hvanyo: I created gap in our jira ticket

marcopagliarulo (approved):

hvanyo (approved):
Merging as preliminary Menu accessibility with known issues & will address accessbility gaps found in another PR.

PR #1029, opened 2021-07-21
[ WIP] - Menu Accessibility Remediation

VancheeZze:

### DIFF
--- a/packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
+++ b/packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
@@ -29,6 +29,7 @@ import {
   designable,
   flowIf,
   not,
+  startWith,
 } from '@bodiless/fclasses';
 
 import { useMenuContext } from './withMenuContext';
@@ -45,6 +46,11 @@ const useHasLink = () => {
   return Boolean(linkHref.data.href);
 };
 
+const ClickOutside = React.forwardRef<any, any>((props, ref) => {
+  useClickOutside(ref as React.MutableRefObject<HTMLLIElement>, props.onClickOutside);
+  return null;
+});
+
 const withSubmenuToggle = (Component: ComponentType<any> | string) => (props: any) => {
   const { activeSubmenu, setActiveSubmenu } = useMenuContext();
   const { node } = useNode();
@@ -56,17 +62,34 @@ const withSubmenuToggle = (Component: ComponentType<any> | string) => (props: an
       : setActiveSubmenu(nodeID)
   );
 
-  // @TODO -- This is not ideal. We need a way to close submenu on click outside
-  // without adding another html element for ref
-  const ref = useRef(null);
-  useClickOutside(ref, () => setActiveSubmenu(undefined));
-
   return (
-    <button type="button" ref={ref} onClick={toggleSubmenu}>
+    <button type="button" onClick={toggleSubmenu}>
       <Component {...props} tabIndex={-1} />
     </button>
   );
 };
+
+const AccessibleMenuItem: FC<any> = props => {
+  const { activeSubmenu, setActiveSubmenu } = useMenuContext();
+  const { node } = useNode();
+  const nodeID = node.path[node.path.length - 2];
+
+  const ref = useRef(null);
+
+  return (
+    <>
+      <li ref={ref} {...props} />
+      {
+        activeSubmenu === nodeID
+          && <ClickOutside ref={ref} onClickOutside={() => setActiveSubmenu(undefined)} />
+      }
+    </>
+  );
+};
+
+const withAccessibleSubmenuItem = withDesign({
+  OuterWrapper: startWith(AccessibleMenuItem),
+});
 
 type SubmenuIndicatorComponents = {
   Button: ComponentType<any>,
@@ -116,7 +139,7 @@ const withMenuNav = asToken(
   withDesign({
     Nav: asToken(
       addClasses('w-full'),
-      addProps({ 'aria-label': 'Main Site Navigation Menu' }),
+      addProps({ 'aria-label': 'Main Site Navigation Menu', role: 'navigation' }),
     ),
   }),
 );
@@ -171,6 +194,7 @@ const withAccessibleSubMenuAttr = withDesign({
  * and accessibility attributes to the submenu items.
  */
 const asAccessibleSubMenu = asToken(
+  withAccessibleSubmenuItem,
   withSubmenuIndicator,
   withAccessibleSubMenuAttr,
 );


packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
```diff
@@ -45,6 +48,59 @@ const useHasLink = () => {
...
+  useClickOutside(ref as React.MutableRefObject<HTMLLIElement>, props.onClickOutside);
+  return null;
+});
+
+const FocusOnEsc = React.forwardRef<any, any>((props, ref) => {
+  if (typeof ref === 'function') return null;
+  if (ref && ref.current) {
+    const escFunction = useCallback((event) => {
+      // Listen for ESC key only
+      if (event.keyCode === 27) {
```
marcopagliarulo, line 61: keyCode is deprecated and could stop to work in future. Despite the wide usage of keyCode probably will force the browsers to keep it working, I'd use it as fallback.
e.g.

```
const key = event.key || event.keyCode
if (key === 'Escape' || key === 'Esc' || key === 27 ) {

```
Note: event.key on IE11 returns Escape

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
```diff
@@ -45,6 +48,59 @@ const useHasLink = () => {
...
+          // Find Button ( List:Title element ) and set focus on it when submenu closes with ESC key
+          if (ref.current.children[i].tagName === 'BUTTON') {
+            ref.current.children[i].focus();
+          }
+        }
+      }
+    }, []);
+
+    useEffect(() => {
+      ref.current.addEventListener('keydown', escFunction, false);
```
marcopagliarulo, line 73: what do you think to use **keyup** instead of **keydown**? Keyup is fired just once, instead on a continuous press keydown is fired many time

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
```diff
@@ -45,6 +48,59 @@ const useHasLink = () => {
...
+  return null;
+});
+
+const asAccessibleMenuTitle = (
+  isSubmenuIndicator: boolean = false,
+) => (Component: ComponentType<any> | string) => (props: any) => {
+  const { node } = useNode();
+  const titleText = node.child<{ text: string }>(DEFAULT_NODE_KEYS.titleNodeKey).data.text;
+  const ariaLabel = isSubmenuIndicator
+    ? `${titleText} - More`
```
marcopagliarulo, line 90: Hardcoded label could not be translated, it would produce a not accessible label on non English pages

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
```diff
@@ -45,6 +48,59 @@ const useHasLink = () => {
...
+
+  return <Component role="menuitem" aria-label={ariaLabel} {...props} />;
+};
+
+const asAccessibleSubMenuTitle = (Component: ComponentType<any> | string) => (props: any) => {
+  const { activeSubmenu } = useMenuContext();
+  const { node } = useNode();
+  const nodeID = node.path[node.path.length - 1];
+
+  return <Component area-haspopup="true" aria-expanded={activeSubmenu === nodeID} {...props} />;
```
marcopagliarulo, line 101: please fix the typo in **aria-haspopup**

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
```diff
@@ -116,7 +198,7 @@ const withMenuNav = asToken(
   withDesign({
     Nav: asToken(
       addClasses('w-full'),
-      addProps({ 'aria-label': 'Main Site Navigation Menu' }),
+      addProps({ 'aria-label': 'Main Site Navigation Menu', role: 'navigation' }),
```
marcopagliarulo, line 201: Defining the role to navigation, a landmark is created, the landmark must be unique, in case of more than one menu with role Navigation, the uniqueness is given by the aria-label, so hardcoding it, two menus on the same pages would not be accessible.

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
```diff
@@ -127,10 +209,15 @@ const withMenuNav = asToken(
 const withAccessibleMenuAttr = withDesign({
   Wrapper: addProps({ role: 'menubar', 'aria-label': 'Navigation Menu' }),
```
marcopagliarulo, line 210: same as for line 201, this is not accessible with more than one menu. Maybe is worth to include this as well in the remediation.

marcopagliarulo (changes requested):

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
VancheeZze: Addressed, Thanks!

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
VancheeZze: Didn't think about it, but makes sense! Updated

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
VancheeZze: Addressed. It is now may be changed by the site builder, but defaults to `More`. Added an example of a custom suffix on the test site.

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
VancheeZze: Good catch! Fixed

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
VancheeZze: I don't think this is an issue. Currently, it is only used inside an `asTopNav()` and it is implied that we are dealing with the Navigation. If site Builder uses two Navigation menus on the same page then I think it is a site builder error.

There are two ways to avoid it. One is to **not** use `asTopNav()` for any menus on the site except for the Main Nav one. The other way is to manually update `aria-label` and `role` attrs on the site level through `addProps()` on appropriate components ( `Nav` and `Wrapper` in this case )

packages/bodiless-navigation/src/Menu/asAccessibleMenu.tsx
VancheeZze: Same as in [this](https://github.com/johnsonandjohnson/Bodiless-JS/pull/1029#discussion_r677752708) comment.

marcopagliarulo (approved):

hvanyo (approved):

PR #1037, opened 2021-08-04
feat(navigation): Menu Supports RTL

VancheeZze:
## Changes
- RTL toggle is added to the menu page

## Test Instructions
- RLT Menu now can be tested on the `/menu` page when RTL toggle is toggled on
- No changes to the menu component. Only the toggle on the test page is added

### DIFF
--- a/examples/test-site/src/components/Card/token.tsx
+++ b/examples/test-site/src/components/Card/token.tsx
@@ -71,7 +71,7 @@ const asCardTextWhite$ = withDesign({
 const asRtlCard$ = withDesign({
   Title: addClasses('rtl:text-right'),
   Body: addClasses('rtl:text-right'),
-})
+});
 
 const asCardHorizontal = withCategory('Orientation')(asCardHorizontal$);
 const asCardVertical = withCategory('Orientation')(asCardVertical$);
--- a/examples/test-site/src/components/Card/token.tsx
+++ b/examples/test-site/src/components/Card/token.tsx
@@ -68,6 +68,11 @@ const asCardTextWhite$ = withDesign({
   Body: addClasses('text-white'),
 });
 
+const asRtlCard$ = withDesign({
+  Title: addClasses('rtl:text-right'),
+  Body: addClasses('rtl:text-right'),
+})
+
 const asCardHorizontal = withCategory('Orientation')(asCardHorizontal$);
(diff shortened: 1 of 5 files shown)


AshuraHassan (approved):

examples/test-site/src/components/Card/token.tsx
```diff
@@ -68,6 +68,11 @@ const asCardTextWhite$ = withDesign({
   Body: addClasses('text-white'),
 });
 
+const asRtlCard$ = withDesign({
+  Title: addClasses('rtl:text-right'),
+  Body: addClasses('rtl:text-right'),
+})
```
Guilherme-Almeida-Zeni, line 74: Hello @VancheeZze ,

Could you please double-check if lint is complaining about missing semicolon and fix it in case of lint errors?

examples/test-site/src/data/pages/menu/index.tsx
```diff
@@ -101,9 +102,24 @@ const Description = addClasses('text-sm mb-2 italic')(Div);
 const DataPreviewContainer = addClasses('overflow-scroll')(Div);
 const DescList = addClasses('list-disc ml-5')(Ul);
 
+const RTLToggle = () => {
+  const [isRtl, setRtl] = React.useState(false);
+  return (
+    <>
+      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
Guilherme-Almeida-Zeni, line 109: Hello @VancheeZze ,

Is it possible to use Label and Input components from @bodiless/fclasses package? This would help to avoid this lint error.

VancheeZze: It is possible, but this linting error is because we do not bind `label` and `input` with ID since `input` is inside the label and there is no need for it. Also it is just for this test page so I don't think it matters if we use these components from fClasses

examples/test-site/src/components/Card/token.tsx
VancheeZze: It does not since it is on the test site level. I propose we do not fix it at this point

Guilherme-Almeida-Zeni (dismissed):

AshuraHassan (approved):

Guilherme-Almeida-Zeni (approved):

PR #1040, opened 2021-08-09
[WIP] Search data layer

allencit:

### DIFF
--- a/packages/bodiless-search/src/components/Search.tsx
+++ b/packages/bodiless-search/src/components/Search.tsx
@@ -55,6 +55,7 @@ type SearchResultComponents = {
   SearchResultList: ComponentType<any>;
   SearchResultListItem: ComponentType<any>;
   SearchResultSummary: ComponentType<StylableProps>;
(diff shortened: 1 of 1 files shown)


packages/bodiless-search/src/components/Search.tsx
```diff
@@ -55,6 +56,7 @@ type SearchResultComponents = {
   SearchResultList: ComponentType<any>;
   SearchResultListItem: ComponentType<any>;
   SearchResultSummary: ComponentType<StylableProps>;
+  GtmHelmet: ComponentType<StylableProps>;
```
dewen, line 59: I don't feel we should include a dependency here, if it's about to add something to head, maybe just add a SearchHelmet wrapper here.

allencit: Hi @dewen, addressed, for this `GtmHelmet`, I did want to add gtm helmet to head, make sense to rename to `SearchHelmet`, could you please check?

konst8 (dismissed):

dewen (approved):

PR #1105, opened 2021-09-23
Tailwind purge whitelist autodiscovery

allencit:


packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/getTailwindConfigs.ts
```diff
@@ -0,0 +1,76 @@
...
+const findTailwindConfigPaths = async () => locateFiles({
+  startingRoot: './',
+  filePattern: new RegExp('^tailwind.config.js$'),
+});
+
+/**
+ * Combination of all available tailwind configs.
+ * @param deps Site level dependencies
+ */
+const getBodilessTailwindConfig = async (deps: string[]) => {
```
wodenx, line 58: Seems this `deps` arg is unused?

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/getTailwindConfigs.ts
```diff
@@ -0,0 +1,76 @@
...
+const getPackageNameFromPackageJson = (
+  packageJsonPath: string,
+): string => getVauleFromPackageJson(packageJsonPath, 'name');
+
+/**
+ * Finds all tailwindcss configuration files.
+ */
+const findTailwindConfigPaths = async () => locateFiles({
+  startingRoot: './',
+  filePattern: new RegExp('^tailwind.config.js$'),
```
wodenx, line 51: I wonder if, instead of finding all `tailwind.config.js` we should require package maintainers to give a special name to those which  they want included/merged -- maybe something like `site.tailwind.config.js`?

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { resolve as resolvePath } from 'path';
+import {
+  getDependenciesFromPackageJson,
+  getBodilessTailwindConfig,
+} from './getTailwindConfigs';
+import { writeToFile } from '../generate-env-vars/utils';
+
+const templateWrap = `
+// This file is generated automatically, please don't change it
```
wodenx, line 24: Ok, but what mechanism would a site-builder use to add site-level tailwind configs?

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    tailwindConfig: require('#pkg/tailwind.config'),
+  }`;
+
+/*
+ * This command will auto discover all tailwind configuration files
+ * and generate the configuration list in the site package
+ */
+const init = async () => {
+  const pkg = resolvePath('package.json');
+  const deps = Object.keys(getDependenciesFromPackageJson(pkg));
```
wodenx, line 39: Do we still need this?  Seems we are walking the node_modules to find them...

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  const cfg = await getBodilessTailwindConfig(deps);
+  const cfgs = cfg.map(item => template.replace(/#pkg/g, item)).join(',');
+  await writeToFile('.tailwindConfigs.js', templateWrap.replace(/#pkgs/g, cfgs));
```
wodenx, line 42: I wonder if we should just write the `tailwind.config.js` and require site-level configs to be in `site.tailwind.config.js` or similar?

wodenx (changes requested):

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/getTailwindConfigs.ts
allencit: It's been used [here](https://github.com/johnsonandjohnson/Bodiless-JS/pull/1105/files#diff-e9bf39c138fc62510005e8dc727fd8b1161fba6db5a4a684b4b2af8bbe5ca7b4R48)

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/getTailwindConfigs.ts
allencit: Addressed as you suggested

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
allencit: Yes, there is two steps to generate the configs automatically.

1.Walking the node_modules to find the pacages which has the site.tailwind.config.js file
2.Make sure the package is be used in the site

That means if a package has the site.tailwind.config.js but the package is not in the site dependencies will not be included

allencit: Yes, there is two steps to generate the configs automatically.

1.Walking the node_modules to find the pacages which has the site.tailwind.config.js file
2.Make sure the package is be used in the site

That means if a package has the site.tailwind.config.js but the package is not in the site dependencies will not be included

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
allencit: Addressed as you suggested

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
allencit: Addressed as you suggested to use site.tailwind.config.js for the site-level one

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
wodenx: hm.... but what about transitive dependencies?  ie - if i depend on `foo` and `foo` depends on `bar` (but i don't require `bar` explicitly), i think i still need the classes uses by `bar`, no?

packages/gatsby-theme-bodiless/src/dist/tailwindcss/getBodilessPackages.ts
```diff
@@ -21,23 +21,23 @@ const getPackageRoot = (packagePath: string) => path.dirname(path.dirname(packag
 const getBodilessPackages = () => [
```
wodenx, line 21: And why do we need this?  Won't we pick up the bodiless packages using he standard mechanism?

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
wodenx: Though I guess there is a problem in that in our monorepo structure, node_modules *may* contain packages which are not even transitive dependencies of the site...  Maybe the right way to resolve this is to print a warning if a "site.tailwind.config.js" file iencountered which is not in a direct dependency, and require that package to be added to package.json.  Or, we could print an error in this case, and require you either to specify the package as a dependency in package.json or to add the package to an explicit exclusion list...

wodenx (changes requested):

packages/gatsby-theme-bodiless/src/dist/generate-tailwind-configs/index.ts
wodenx: Actually - will the `require` fail if the package is not listed in the site level package.json?  Also - we probably will often want to run this using the dependencies of a brand package, not the site itself, so it would be good to have a way to specify one (or more) packages whose dependencies should be searched...

PR #1106, opened 2021-09-23
[WIP] Disable menu

konst8:
## Changes
Allow users to disable the page menu independently


packages/bodiless-components/src/PageDisableButton/index.tsx
```diff
@@ -58,115 +49,157 @@ type Data = {
...
-          },
-        },
-      });
-    }
+    }, []);
+
+    return (
+      <>
+        {Object.entries(formValues).map(
+          ([key, value]) => <h1 key={key}>{`${key}: ${value ? 'Disabled' : 'Enabled'}`}</h1>,
```
VancheeZze, line 164: Should it be `<h1>`? I'd suggest either `<h3>` if it has to be heading or maybe a list items `<li>` inside the parent `<ul>` for all of the list items.

packages/bodiless-components/src/PageDisableButton/index.tsx
```diff
@@ -58,115 +49,157 @@ type Data = {
...
-    };
-    return <Form {...props} />;
-  };
-  const { buttonLabel, formIcon } = formState;
+  return (
+    <>
+      <ComponentFormTitle>
+        Disable Status
+      </ComponentFormTitle>
+      {step < 1 && <StepOneContent />}
```
VancheeZze, line 175: This may be a little confusing, I'd suggest we check for the step explicitly:
```
{step === 0 && <StepOneContent />}
{step === 1 && <StepTwoContent />}
```

packages/bodiless-components/src/PageDisableButton/index.tsx
```diff
@@ -58,115 +49,157 @@ type Data = {
...
-  const context = useEditContext();
-  const isPageDisabled = useIsPageDisabled();
+  const toggleOffPageCheckbox = () => {
+    setValue('Page', false);
+  };
 
-  const togglePageVisibility = (): void => {
-    if (isPageDisabled) {
-      // Enable
+  const StepOneContent = useCallback(() => {
```
VancheeZze, line 100: Just thinking out loud -- should we rename `StepOneContent` and `StepTwoContent` into `FeaturesSelectForm` and `ConfirmationForm`?

If we do that, we can update `setStep(1)` to accept a string. Something like
```
// Create an enum with the step values
enum Steps = { FeaturesSelect, Confirmation }

// Then in a place where we initialize `step` we would do something like
setStep(Steps.FeaturesSelect)

// and on first submit
setStep(Steps.Confirmation)

// With all that we can check what form to render by
{step === Steps.FeaturesSelect && <FeaturesSelectForm />}
{step === Steps.Confirmation && <ConfirmationForm />}
```

Not necessarily to implement, just interesting what do you think? It doesn't change functionality, just for the readability.

packages/bodiless-navigation/src/Menu/MenuTitles.tsx
```diff
@@ -63,12 +65,31 @@ const asMenuLink = (asEditableLink: typeof asBodilessLink) => asToken(
   ),
 );
 
+/**
+ * Token that check if the current menu item is disabled for static and preview mode.
+ * Returns an empty fragment if the link is listed in src/data/site/disabled-pages.json
+ */
+const asDisableableMenuLink: Token = Component => props => {
```
VancheeZze, line 72: How do you like the `asDisabledPageLink` name for this HOC?

packages/bodiless-components/src/PageDisableButton/index.tsx
konst8: Done.

packages/bodiless-components/src/PageDisableButton/index.tsx
konst8: Done.

packages/bodiless-components/src/PageDisableButton/index.tsx
konst8: Yes, I like it. Implemented, thanks.

VancheeZze (approved):

konst8:
Declined in favor of https://github.com/johnsonandjohnson/Bodiless-JS/pull/1143

Issue #1147, opened 2021-10-12
Bodiless OIDC


hvanyo:
### User story
- As a site owner, I want to allow users to establish identity and enter basic profile information, So that I can collect information about my users. (Registration as default separate url e.g. login.example.com)
- As a site owner, I want to make sure openID token which is signed with signature is validated at any interaction, So that the security of the system and customer's data is ensured

## Acceptance Criteria
AC0: There is a **OIDC** Package in Bodiless

AC1: There is reusable client side API that utilizes OIDC for authentication 

AC2: There is a component which provides a login link and when clicked invokes the **TEST** login flow and stores the tokens in a place accessible to other components. 

AC3: The Login component accepts a prop "prompt", the allowed values are "login".  login is the default value

AC4: Functional unit tests are provided as needed.

### *Prerequisites*
- Access & use sandbox in janrain sandbox (Sergeii & Marco to get creds)

### *Documentation*

**Akamai/Janrain Hosted Login Documentation:**

https://janrain-education-center.knowledgeowl.com/home/end-user-screens

**User experience:**
- SIgn In: https://janrain-education-center.knowledgeowl.com/home/end-user-screens#signin
- Registration: https://janrain-education-center.knowledgeowl.com/home/end-user-screens#registration
- Forget Password: https://janrain-education-center.knowledgeowl.com/home/end-user-screens#forgot
- User Profile: https://janrain-education-center.knowledgeowl.com/home/end-user-screens#profile

### ACs descriptions:
- Configure redirect URIs (redirect URI(s) defines where the user lands after successful login or registration (https://identitydocs.akamai.com/getting-started/hosted-login/step3/page1-redirecturis/)
  - Get a token to use in configuration calls
  - Configure your redirect URI(s)
- Hosted Widget integration is done using the recommended by Akamai third-party library/service:
  - Option 1 Utilize OIDC https://github.com/thchia/react-oidc
    - Build UX elements that trigger an authentication request to the Identity Cloud.
    - Implement OpenID Connect in order to receive identities from the Identity Cloud.
    - Perform additional authorization checks required by your application.
    - Update the application session with the user information and determine when users must re-authenticate.
    - Logout and revoke issued tokens.
    - DOCs https://identitydocs.akamai.com/home/integrating-oidc-apps-with-the-identity-cloud
    - **_Possible Suggested temporary solution for this ticket due to delivery time (early Nov)_**
  - Option 2 
[...]

PR #1150, opened 2021-10-13
feat(oidc): Akamai Hosted Login

VancheeZze:
## Changes
- New `@bodiless/oidc` package.

## Test Instructions
- Try to Log In and Log Out using the button on the `/oidc` page.
- You can use `bob/bob` or `alice/alice` for login/password

## Related Issues
Closes: https://github.com/johnsonandjohnson/Bodiless-JS/issues/1147

## Notes
- Currently uses https://demo.identityserver.io/ as an Identity Provider.


wodenx:
per discussions i think this can be constructed so as to be agnostic as to the OIDC provider, so i think we should rename to bodiless/oidc.  cc @marcopagliarulo  @VancheeZze

tsconfig.settings.json
```diff
@@ -47,6 +47,8 @@
       "@bodiless/layouts-ui/*": ["bodiless-layouts-ui/src/*"],
       "@bodiless/navigation": ["bodiless-navigation/src"],
       "@bodiless/navigation/*": ["bodiless-navigation/src/*"],
+      "@bodiless/oicd": ["bodiless-oicd/src"],
```
awm086, line 50: oidc?

packages/bodiless-oidc/src/AuthContext.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * @param args OIDC Sign In arguments
+ * @see SigninRedirectArgs
+ */
+export const withSignInOnClick = (
+  args?: SigninRedirectArgs,
+) => (Component: ComponentType) => (props: any) => {
+  // Property 'signIn' does not exist on type 'AuthContextProps | undefined'
+  // @ts-ignore
+  const { signIn } = useBodilessOidc();
+  return <Component {...props} onClick={() => signIn(args)} />;
```
marcopagliarulo, line 46: What do you think to use a conditional return in base of signIn null? Same applied below

VancheeZze:
Closing this PR in favor of https://github.com/johnsonandjohnson/Bodiless-JS/pull/1190 created to the "main" branch.

packages/bodiless-oidc/src/AuthContext.tsx
VancheeZze: `signIn` and `signOut` will always be there in this case. Those come with the provider and are always present. Added a type cast `as AuthContextProps`

PR #1184, opened 2021-10-30
fix(backend) Fix git tests.

wodenx:
## Changes
- Update git fixture for backend tests to have a "main" branch


wodenx:
Wrong targetr branch

Issue #1165, opened 2021-10-25
BodilessJS transpiles to ES2015


marcopagliarulo:
## User Story

- AS A Developer
- I WANT BodilessJS Packages compiled into ES2015 instead of commonJS
- SO THAT tree shaking works on JS Bundle generation

### Background/Description

Webpack is able to remove dead code from the js bundle relying on the static structure of ES2015 module syntax. 
BodilessJS is transpiled to CommonJS, preventing Webpack to perform tree shaking.
 
### Acceptance Criteria

AC1. BodilessJS is transpiled to ES2015
AC2. A rough evaluation of side effects is performed and reflected on the package.json files 
* focus on higher order components, instead of file by file to check everything to see if it modifies any global context 
* spend approx 10/15 mins evaluating each package

AC3. There are no security alerts -> https://github.com/johnsonandjohnson/Bodiless-JS/security

Testing Note: Regression Testing of the site components 

### Implementation Notes
POC: https://github.com/marcopagliarulo/Bodiless-JS/commit/a8b21c30838b71d5fa0f751e7f101a7bcf719523

Note: gatsby-theme-bodiless cannot use ES2015 because the JS files using Gatsby APIs need to be commonjs.

### References
https://webpack.js.org/guides/tree-shaking

PR #1208, opened 2021-11-05
feat: Add tree shaking support to all browser packages

lucas-varela:
## Changes
This PR adds tree shaking support to all packages that may be used in a browser.

The following packages are only used on Node and **won't** support tree shaking:
-  @bodiless/backend
-  @bodiless/cli
-  @bodiless/documentation
-  @bodiless/migration-tool
-  @bodiless/psh
-  @bodiless/gatsby-plugin-ssi
-  @bodiless/headless-chrome-crawler

Note on html2canvas: this package was used with a dynamic import in the past to prevent breaking SSR. The package has since been updated to support SSR, so I removed the dynamic import as it is no longer necessary.

## Related Issues
This PR closes https://github.com/johnsonandjohnson/Bodiless-JS/issues/1165


packages/bodiless-layouts/src/ComponentSelector/ScreenShot.tsx
```diff
@@ -13,11 +13,10 @@
  */
 
 import React from 'react';
+import html2canvas from 'html2canvas';
```
marcopagliarulo, line 16: The import of html2canvas was limited to the frontend, could this import change create issue during SSR?

marcopagliarulo (changes requested):
I think we are missing some required changes.

the typescript configuration is still set to transpile to commonjs
https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/tsconfig.settings.json#L5-L6

target and module should be set to ES2015

I think we should use in https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/babel.config.js

  "presets": [
    [
      "@babel/preset-react",
      { modules: false }
    ]
  ]

packages/bodiless-layouts/src/ComponentSelector/ScreenShot.tsx
lucas-varela: Not anymore, the package has since been updated to support SSR. It still doesn't work on node, the difference now is that it simply won't do anything, instead of throwing an error. Since the function that uses this package only ever runs on browsers anyway, it's safe to load it here statically.

babel.config.js
```diff
@@ -1,6 +1,15 @@
...
+  'plugins': [
+    ['@babel/plugin-proposal-decorators', { 'legacy': true }],
+    ['@babel/plugin-proposal-class-properties', { 'loose': true }],
+  ],
+  'env': {
+    'test': {
+      'plugins': [
+        ['@babel/plugin-proposal-decorators', { 'legacy': true }],
+        ['@babel/plugin-proposal-class-properties', { 'loose': true }],
+        ['@babel/plugin-transform-modules-commonjs'],
```
marcopagliarulo, line 11: since we discovered the jest issue was due to the target set to es2015, I think we can remove this plugin as not useful anymore

package.json
```diff
@@ -69,6 +69,7 @@
     "@babel/core": "^7.9.0",
     "@babel/plugin-proposal-class-properties": "^7.2.1",
     "@babel/plugin-proposal-decorators": "^7.4.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.5.0",
```
marcopagliarulo, line 72: as by comment above

packages/bodiless-documentation/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../../tsconfig.settings.json",
   "compilerOptions": {
+    "target": "es5",
```
marcopagliarulo, line 4: you don't need to override the target here since the value is the same of the base config

packages/bodiless-search/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../../tsconfig.settings.json",
   "compilerOptions": {
+    "target": "es5",
```
marcopagliarulo, line 4: you don't need to override the target here since the value is the same of the base config

packages/gatsby-theme-bodiless/tsconfig.json
```diff
@@ -1,6 +1,8 @@
 {
   "extends": "../../tsconfig.settings.json",
   "compilerOptions": {
+    "target": "es5",
```
marcopagliarulo, line 4: you don't need to override the target here since the value is the same of the base config

babel.config.js
lucas-varela: Although our code is being compiled to es5, we're still using es2015 modules, so this plugin is necessary. I removed the other two, however, since it's redundant to add them to this list.

package.json
lucas-varela: See above

marcopagliarulo (dismissed):

marcopagliarulo (approved):

lucas-varela:
New PR using local branch: #1251

PR #1246, opened 2021-11-19
fix(oidc): Fix new oidc-client-ts types breaking build

### LINKED ISSUE #1243: Upgrade OIDC package and fix new type errors
## Description
Downstream sites are hosting the oidc-client-ts v2 package and produce the following error
## Issue
``
    
    
      @bodiless/oidc: src/AuthCallback.tsx(49,23): error TS2345: Argument of type 'void | User' is not assignable to parameter of type 'SetStateAction<User | null>'.
@bodiless/oidc:   Type 'void' is not assignable to type 'SetStateAction<User | null>'.
@bodiless/oidc: src/AuthCallback.tsx(50,34): error TS2345: Argument of type 'void | User' is not assignable to parameter of type 'User | null'.
@bodiless/oidc:   Type 'void' is not assignable to type 'User | null'.
@bodiless/oidc: src/AuthCallback.tsx(51,36): error TS2345: Argument of type 'void | User' is not assignable to parameter of type 'User | null'.
lerna ERR! npm run build exited 2 in '@bodiless/oidc'
``

lucas-varela:
## Changes

Fixed `oidc-client-ts` package breaking our builds after making a breaking change on v2.0.0-rc.

It's important to note that, even though this type should not break even after new updates, this package is still on pre-release and new breaking changes might be introduced in the future, specially because it is not following SemVer as commonly expected.

It might be a good idea to pin its version while a stable release isn't out.

## Related Issues

This fixes #1243


packages/bodiless-oidc/src/types.ts
```diff
@@ -107,7 +107,7 @@ export type AuthProviderProps = {
    *
    * defaults to 'location=no,toolbar=no,width=500,height=500,left=100,top=100'
    */
-  popupWindowFeatures?: string;
+  popupWindowFeatures?: UserManagerSettings['popupWindowFeatures'];
```
dewen, line 110: why not import type PopupWindowFeatures? Also can we update the comment since it's no longer default to string.

dewen: I guess UserManagerSettings['popupWindowFeatures'] implicit type is more flexible.

lucas-varela:
Already fixed in the scope of #1251

dewen (approved):

konst8 (approved):

PR #1251, opened 2021-11-23
feat: Add tree shaking support to all browser packages

lucas-varela:
## Changes
This PR adds tree shaking support to all packages that may be used in a browser.

The following packages are only used on Node and **won't** support tree shaking:
-  @bodiless/backend
-  @bodiless/cli
-  @bodiless/documentation
-  @bodiless/migration-tool
-  @bodiless/psh
-  @bodiless/gatsby-plugin-ssi
-  @bodiless/headless-chrome-crawler

Note on html2canvas: this package was used with a dynamic import in the past to prevent breaking SSR. The package has since been updated to support SSR, so I removed the dynamic import as it is no longer necessary.

## Related Issues
This PR closes https://github.com/johnsonandjohnson/Bodiless-JS/issues/1165


lucas-varela:
New PR using local branch on #1251

examples/starter/package.json
```diff
@@ -38,7 +38,6 @@
     "express": "^4.17.1",
     "gatsby": "~3.13.0",
     "gatsby-plugin-canonical-urls": "^3.6.0",
-    "gatsby-plugin-google-tagmanager": "^3.6.0",
```
marcopagliarulo, line 41: why are you removing gatsby-plugin-google-tagmanager in scope of this story?

packages/bodiless-search/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.settings.json",
   "compilerOptions": {
+    "module": "commonjs",
```
marcopagliarulo, line 4: why are you overriding the module for this package?

examples/starter/package.json
lucas-varela: I must have deleted this line accidentally, my bad. Will re-add it.

packages/bodiless-search/tsconfig.json
lucas-varela: This package exports a node script called `create-search-index`, which can't use ES modules. There's only one alternative to using commonjs, which is running two separate builds on this package, one using ES modules on most files, and the other using commonjs only on the node script. If you think this alternative is viable, I can implement it.

marcopagliarulo: We have to find a similar solution for gatsby-bodiless-theme, so I'd group these same issues together

marcopagliarulo (approved):

examples/starter/package.json
TatsianaMiraniuk: @hvanyo , @lucas-varela  - can you please clarify why GTM has been deleted?

hvanyo: @TatsianaMiraniuk. Current issues is we have dependency mismatch between bodiless/canvasx.   The starter/test site doesn't have GTM configured correctly (it runs with XXXXX number) and GTM in the past has been deferred from being tested on test site as most of the GTM activation is within Canvasx.     Right now we have problem as Canvasx GTM packages uses a later version and Bodiless is at older version.   We had 3 choices:  upgrade dependency on bodiless, but its not tested here  OR  remove it and just keep it at canvasx level where it is tested OR downgrade Canvasx but that is going backward.   chris and I decided the former was better cause otherwise it requires us always updating bodiless for artificial dependency.

TatsianaMiraniuk: @hvanyo thanks!

hvanyo (dismissed):

marcopagliarulo (dismissed):

hvanyo (approved):

TatsianaMiraniuk (approved):

Issue #1236, opened 2021-11-17
Improve component selector with large number of components


hvanyo:
## Description
As a content editor
I want to be able to select from a large number of components in a flow container or chameleon
Without severely impacting page performance for myself or end users.

## Acceptance Criteria
AC1.
The design passed to a flow container or chameleon  is only partially applied when the flow container or chameleon is mounted (only those keys which are actually rendered on the page are processed) AND the first 25 that appear in the component selector are rendered

AC2. When the user selects the component selector the remaining components are processed. 

AC3.
For a flow container or chameleon which uses the component selector, the entire design passed through  to the component selector, which applies it in order to render the options,

AC4.
For a chameleon which does not use the component selector, the design is passed through to the form, but is not applied -- only the keys of the design are used to create the form.

AC5. Test page(s) are added to the Bodiless Test site. One is for a  large number of components in a flow container and the other is a page for a large number of chameleon variations.  (Edited on 1/20/2022 by hvanyo to remove word 'Two' test pages)
AND
The page load time and interaction of the edit interface is not impacted

PR #1259, opened 2021-12-02
chore(performance): Mitigate page weight

konst8:
## Changes
Create FC and Chameleon selectable components only when CS form is opened.

## Related issues
https://github.com/johnsonandjohnson/Bodiless-JS/issues/1236


packages/bodiless-components/src/Chameleon/withChameleonButton.tsx
```diff
@@ -157,8 +159,8 @@ const withChameleonButton = <P extends object, D extends object>(
   const useMenuOptions = (props: P & EditButtonProps<D>) => {
     const overrides = useOverrides(props);
     // if useOverrides returns falsy, it means not to provide the button.
-    const { selectableComponents } = useChameleonContext();
-    const extMenuOptions = Object.keys(selectableComponents).length > 1
+    const { design } = useChameleonContext();
```
wodenx, line 162: And here - we need to omit default key in making this choice.  In all cases, you need to apply the design to get the metadata - so this is sightly problematic.  One way to resolve - check the number of keys in the design - if it's <= 2 then apply the design to get the metadata, and make the choice (between swap button and toggle button) based on whether the default key has a title.  Otherwise, defer applying the design (and filtering out the default key) to later.

wodenx (changes requested):

packages/bodiless-components/src/Chameleon/withChameleonContext.tsx
```diff
@@ -55,48 +55,65 @@ const useChameleonContext = (): ChameleonState => {
...
-        />
-      </ChameleonContext.Provider>
-    );
-  };
+  const WithChameleonContext: FC<any> = props => (
+    <ChameleonContext.Provider value={{
+      RootComponent: RootComponent,
+      isOn: getIsOn(props),
+      activeComponent: getActiveComponent(props),
+      components: props.components,
```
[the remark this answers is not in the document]
wodenx: I see - you only do it if the selector can be toggled - but I think it's confusing.  Really, all you need is to know whether the default key has a title, and the only place you need it is in `withChameleonButton` to determine what button to show -- so probably a better solution would be to check that key only there, like
```
const useShowToggleButton = () => {
  const { design } = useChameleonContext();
  if (Object.keys(design).length < 2) return true;
  if (Object.keys(desigh).length  > 2) return false;
  if (desigh[DEFAULT_KEY) === undefined) return false;
  const test = design[DEFAULT_KEY)(Fragment);
  return test.title === undefined;
}
```

packages/bodiless-components/src/Chameleon/withChameleonContext.tsx
```diff
@@ -55,48 +55,65 @@ const useChameleonContext = (): ChameleonState => {
...
+  if (Object.keys(design).length <= 2) {
+    return true;
+  }
+  return false;
+};
 
 const withChameleonContext = (
   nodeKeys: WithNodeKeyProps,
   defaultData?: ChameleonData,
+  RootComponent: ComponentOrTag<any> = Fragment,
```
wodenx, line 94: As discussed, either use this also when applying the chameleon or supply it as an argument to 'useChameleonSelectorForm`.  In either case, we want to be sure that `asBodilessChameleon` still gets it from the target, eg
```
asBodilessChameleon('nodeky')(ComponentX)
```
makes ComponentX the RootComponent.

wodenx (changes requested):

wodenx (dismissed):

packages/bodiless-components/__tests__/asBodilessChameleon.test.tsx
```diff
@@ -306,24 +313,5 @@ describe('asBodilessChameleon', () => {
       Baz: identity,
       [DEFAULT_KEY]: withTitle('DefaultTitle'),
     });
-
-    it('Applies a design correctly', () => {
```
wodenx, line 310: I don't think completely disabling this test is right - instead i think you should just rewrite the assertion so that it passes regardless of the order of the keys.

packages/bodiless-layouts/src/FlowContainer/index.tsx
```diff
@@ -14,7 +14,7 @@
 
 import React, { FC, ComponentType } from 'react';
 import { observer } from 'mobx-react-lite';
-import flow from 'lodash/flow';
+import { flow, uniq, pick, omit } from 'lodash';
```
hvanyo, line 17: Q is this appropriate or do we want to four imports -- lodash/flow, lodash/uniq ?

wodenx:
Replaced by #1308

Issue #1198, opened 2021-11-04
Move Assets on Moved paged


hvanyo:
## Description
As a Site Builder, I want the assets to be move when the user moves the page, so that any assets are moved into the new namespace.

## Acceptance Criteria 
AC1.

When A page is moved that has uploaded assets 
Each asset is moved into the namespace of the moved page and the data is updated to point to those assets.

Note: The assets on the source page are assumed to be already in the correct namespace.

**DUE to technical limitations encountered during implementation** we are 'copying assets to new location and data is updated to point to these assets" and we will purge assets on build in this ticket: https://github.com/johnsonandjohnson/Bodiless-JS/issues/1348

asiarheyeu:
Fixed with #1326

PR #1296, opened 2021-12-14
Add functionality to move assets when move the page

dasdib:
## Changes

- A feature that enable when a page is moved, related assets will move as well.

## Related Issues
Issue [#1198](https://github.com/johnsonandjohnson/Bodiless-JS/issues/1198)

### DIFF
--- a/packages/bodiless-backend/src/page.js
+++ b/packages/bodiless-backend/src/page.js
@@ -406,6 +406,17 @@ class Page {
     });
     return readPromise;
   }
+
+  removePageAssets(path) {
+    return new Promise((resolve, reject) => {
+      fse.remove(path, err => {
+        if (err) {
+          reject(err);
+        }
+        resolve();
+      });
+    });
+  }
 }
 
 module.exports = Page;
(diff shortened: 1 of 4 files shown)


rageshchalil (dismissed):

packages/bodiless-backend/src/page.js
```diff
@@ -314,10 +314,27 @@ class Page {
...
+                let fromPattern = originPath;
+                let toPattern = destinationPath;
+
+                if (path.extname(item) === '.json') {
+                  try {
+                    const Obj = await fse.readJson(fileToBeUpdated);
+                    let fromPath = fromPattern.replace(/\//g, '\\');
+
+                    if (Obj.src && Obj.src.includes(fromPath)) {
+                      toPattern = toPattern.replace(/\//g, '\\\\');
```
konst8, line 326: @dasdib why are these replacements?

dasdib: @konst8 It is to handle specific cases for windows, if you add any images from windows OS, it will contain double backslash in its path. Now during clone ops, you need to replace such origin path to destination path, we need this replacement.

konst8: @dasdib I see now, why it was necessary, thanks.

konst8:
Moved to https://github.com/johnsonandjohnson/Bodiless-JS/pull/1326

PR #1307, opened 2021-12-15
fix(adminui): Fix admin ui issue

allencit:

### DIFF
--- a/packages/bodiless-ui/package.json
+++ b/packages/bodiless-ui/package.json
@@ -6,8 +6,7 @@
   "license": "Apache-2.0",
   "main": "lib/index.js",
   "sideEffects": [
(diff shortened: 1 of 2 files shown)


packages/bodiless-ui/package.json
```diff
@@ -6,8 +6,8 @@
   "license": "Apache-2.0",
   "main": "lib/index.js",
   "sideEffects": [
-    "./src/bodiless.index.css",
```
wodenx, line 9: I think we need the ./src versions also so that it works in the monorepo

allencit: Fixed, thanks

hvanyo (approved):

PR #1323, opened 2021-12-23
Fix lodash import on withMovePageButton.tsx

rageshchalil:
Fix lodash import on withMovePageButton.tsx

### DIFF
--- a/packages/gatsby-theme-bodiless/src/dist/withMovePageButton.tsx
+++ b/packages/gatsby-theme-bodiless/src/dist/withMovePageButton.tsx
@@ -20,7 +20,7 @@ import React, {
   ComponentType,
   HTMLProps,
 } from 'react';
-import { flow } from 'lodash';
+import flow from 'lodash/flow';
 import {
   contextMenuForm,
   useMenuOptionUI,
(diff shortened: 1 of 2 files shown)


packages/gatsby-theme-bodiless/src/dist/withMovePageButton.tsx
```diff
@@ -20,7 +20,7 @@ import React, {
   ComponentType,
   HTMLProps,
 } from 'react';
-import { flow } from 'lodash';
+import flow from 'lodash/flow';
```
konst8, line 23: Hi @rageshchalil 
I think it makes sense to optimize the lodash imports in the examples folder as well, e.g.:
https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/examples/test-site/src/components/YouTube/index.tsx#L15

konst8 (changes requested):

packages/gatsby-theme-bodiless/src/dist/withMovePageButton.tsx
rageshchalil: @konst8 
This is resolved now

konst8 (approved):

hvanyo (approved):

PR #1328, opened 2022-01-05
fix(gatsby-theme-bodiless): Fix changes on bodiless packages not triggering rebuilds

lucas-varela:
## Changes

This fixes the problem which made Webpack not trigger new builds when changing files outside the current site folder, eg. a package file.

Webpack 5 implemented a new optimization to ignore "immutable" folders when watching for changes. By default, everything inside `node_modules` is ignored, since files inside this folder don't normally change. This overrides the default by ignoring all folders inside `node_modules`, except the `@bodiless` folder.

## Test Instructions

Start a test-site or starter server, wait for the build to finish and change any TS file inside any Bodiless package. The server should trigger a new build automatically, as expected.

### DIFF
--- a/packages/gatsby-theme-bodiless/gatsby-node.js
+++ b/packages/gatsby-theme-bodiless/gatsby-node.js
@@ -281,20 +281,4 @@ exports.onCreateWebpackConfig = (
       },
     });
   }
-
-  if (stage === 'develop') {
-    actions.setWebpackConfig({
-      // On development, we want changes on Bodiless packages to trigger
-      // new builds. Webpack won't watch packages inside node_modules by
-      // default, so we remove the @bodiless folder from its default list.
-      //
-      // See: https://webpack.js.org/configuration/other-options/#snapshot
-      snapshot: {
-        managedPaths: glob.sync(
-          './node_modules/!(@bodiless)*', 
-          { absolute: true }
-        ),
-      }
-    });
-  }
 };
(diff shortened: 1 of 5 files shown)


wodenx (dismissed):

packages/gatsby-theme-bodiless/gatsby-node.js
```diff
@@ -282,4 +281,20 @@ exports.onCreateWebpackConfig = (
       },
     });
   }
+
+  if (stage === 'develop') {
```
wodenx, line 285: I"m trying to decide whether we want this in gatsby-theme-bodiless or at site level (and in canvasx starter).  If this is only an issue when using lerna monorepos, then maybe we don't want to be opinionated about it, since someone might use bodiless in a single-site repo, and not want to watch bodiless.  
Also - the paths to manage might be different for different sites - eg for canvasx we probably want to add @canvasx...

packages/gatsby-theme-bodiless/gatsby-node.js
```diff
@@ -20,13 +20,12 @@
...
 const pathUtil = require('path');
 const fs = require('fs');
-
 const { getDisabledPages } = require('@bodiless/components/node-api');
 const { createFilePath } = require('gatsby-source-filesystem');
 const { onCreateNode, createSlug } = require('./create-node');
 const createRedirectAlias = require('./create-redirect-alias');
-
 const Logger = require('./Logger');
+const glob = require('glob');
```
wodenx, line 28: Do we still need this import?

lucas-varela: Yeah, we're using glob on line 39.

lucas-varela: Sorry, my mistake. Removed this line.

wodenx (approved):

Guilherme-Almeida-Zeni (approved):

Issue #1122, opened 2021-09-28
Ability edit redirects


hvanyo:
## Description
As an end user I want ability to edit the existing redirects so I can modify them and update them as required.

AC1.
WHEN the user is in the Edit UI with Edit enabled and selects TOOLS
THEN there is a submenu with "Aliases"  
and when they click on Aliases
They are presented with a panel that contains an editable text of redirects from the 'Redirect Store' in the following format:
* A set of redirects per row
* Each row content is spaced delimited.
* The row includes URL_FROM, URL_TO, & STATUSCODE
and there is check at bottom of form.

AC2.
Upon clicking check at bottom, the edited file is validated.
- If file is validated, the 'Redirect store' is updated and message is provided to "Redirect aliases file validated and saved."
- If not validated, a validatation error message (italic) is provided to end user "The redirects are not valid, please correct."
- **If the line is entered with only two paths and without a number, then default & apply a STATUSCODE of 301 to the line upon save**

Validation:
* check for path as the first, url or path as second and number for third per line.
* no need to validate the urls or path exist

AC3.
Any redirects storied in the 'Redirect store' redirect as expected.
 

**UX**
* https://jhdfdc.axshare.com/#g=1&p=generic_ui&id=sxj4x9

**Implementation Notes**

hvanyo:
@kbray0009 

Here is example of redirects

`
example/producta example/productb 301
example/article/one /example/article/two 301
example/campaign/special / 301
example/contact-us /contact-us 301
example/article/subcategory_one/howto  example/article/subcategory_two/howto 301
`

PR #1293, opened 2021-12-10
feat(redirect): Ability edit redirects

allencit:

PR #1330, opened 2022-01-06
feat(redirect): Ability edit redirects

Guilherme-Almeida-Zeni:
## Overview
New feature to allow user edit redirects through form edition. The redirect mechanism is not part of the scope, it has been developed on #1272.

## Changes
- New form for redirect on bodiless components package
- Input validation rules implemented as required on #1122
- Implemented default size token for form panels.
- Removed resizer from textarea fields globally.


packages/bodiless-components/src/RedirectAlias/form.tsx
```diff
@@ -0,0 +1,248 @@
...
+const ALIASPARTSCOUNT = 3;
+
+const isTextEmpty = (text: string) => (!text || text === '');
+
+const isTextValidate = (text: string): boolean => {
+  // Users must be able to save no redirects.
+  if (isTextEmpty(text)) return true;
+
+  // Any double whitespaces are invalid.
+  if (text.indexOf('  ') >= 0) return false;
```
lucas-varela, line 59: Maybe you should check for 2 or more spaces, just in case. You can try to match a `\ {2,}` regex or something like it.

packages/bodiless-components/src/RedirectAlias/form.tsx
```diff
@@ -0,0 +1,248 @@
...
+
+const REDIRECT_ALIASES = 'Redirect Aliases';
+const CONFIRMATION = 'Redirect aliases file validated and saved.';
+const INVALIDATED = 'The redirects are not valid, please correct.';
+const DEFAULT_REDIRECT_STATUS = '301';
+const ALIASPARTSCOUNT = 3;
+
+const isTextEmpty = (text: string) => (!text || text === '');
+
+const isTextValidate = (text: string): boolean => {
```
lucas-varela, line 54: It should be `isTextValid`.

packages/bodiless-components/src/RedirectAlias/form.tsx
```diff
@@ -0,0 +1,248 @@
...
+    node.setData(convertAliasTextToJson(aliases as string));
+    setStep(Steps.Confirmation);
+  };
+
+  const EditForm = useCallback(() => {
+    useEffect(() => {
+      // Gets initial values from node.
+      const values = {
+        aliases: initialAliases,
+        isValidate: true,
```
lucas-varela, line 162: It should be `isValid`.

packages/bodiless-components/src/RedirectAlias/form.tsx
```diff
@@ -0,0 +1,248 @@
...
+  if (isTextEmpty(text)) return true;
+
+  // Any double whitespaces are invalid.
+  if (text.indexOf('  ') >= 0) return false;
+
+  try {
+    const aliases = text.split('\n');
+    const validatedAliases = aliases.filter(item => {
+      // For each item, first and last chars must not be whitespace.
+      if (item[0] == ' ' || item[item.length - 1] == ' ') {
```
lucas-varela, line 65: In case one or more lines have whitespaces on either end, it may be hard for the user to find the issue, since the invalid error description is generic.

A suggestion: map through each line and use `string.trim()` to automatically deal with whitespace. In case all lines are valid, return the trimmed aliases on the `isTextValid` function instead of true. You could then pass the trimmed aliases directly to the `convertAliasTextToJson`, and the user wouldn't need to deal with this problem themselves.

packages/bodiless-components/src/RedirectAlias/form.tsx
```diff
@@ -0,0 +1,248 @@
...
+      ) {
+        return false;
+      }
+
+      return true;
+    });
+
+    return validatedAliases.length === aliases.length;
+  } catch (error) {
+    return false;
```
lucas-varela, line 89: Could this cause silent errors? The only possible error I can foresee happening here is accessing an out-of-bounds array index, but maybe we should at least log the error in the console, just in case. Feel free to ignore this one, though.

packages/bodiless-components/src/RedirectAlias/form.tsx
Guilherme-Almeida-Zeni: Hello @lucas-varela ,

Actually, this wouldn't be needed. By induction, all cases with 3 or more spaces would be matched if you only look for two, since n = 2 -> true, n+1 -> true, n+2 -> true ... n+i -> true. However, I adjusted as you suggested anyway. Please, check it again.

packages/bodiless-components/src/RedirectAlias/form.tsx
Guilherme-Almeida-Zeni: Hi @lucas-varela ,

Updated. Please, check it again.

packages/bodiless-components/src/RedirectAlias/form.tsx
Guilherme-Almeida-Zeni: Hi @lucas-varela ,

Updated. Please, check it again.

packages/bodiless-components/src/RedirectAlias/form.tsx
Guilherme-Almeida-Zeni: Hello @lucas-varela ,

Actually, first implementation from Allen handled the strings as you suggested but bugs were reported. So, instead of trimming it, we are intentionally checking if there are whitespaces at the beginning and at the end on purpose, to match agreed requirements.

packages/bodiless-components/src/RedirectAlias/form.tsx
Guilherme-Almeida-Zeni: Hello @lucas-varela ,

I don't think so since it will be triggered anyways as invalid message for the admin user because, probably in this case, the expected pattern will not be followed (which can cause the example you mentioned about out-of-bounds). However, I think it's not harmful to put a console error there so I updated it.

lucas-varela (approved):

examples/test-site/src/data/site/disabled-pages.json
```diff
@@ -0,0 +1,3 @@
+{
```
hvanyo, line 1: @Guilherme-Almeida-Zeni  I am not a really fan of adding this by default.  I think the site should work without it.   We have a ticket to do so.   Is there a reason why you added it?

packages/bodiless-components/src/RedirectAlias/form.tsx
```diff
@@ -0,0 +1,258 @@
...
+      ) {
+        return false;
+      }
+
+      return true;
+    });
+
+    return validatedAliases.length === aliases.length;
+  } catch (error) {
+    console.error(error);
```
hvanyo, line 95: @Guilherme-Almeida-Zeni do we want to keep this as a console error or push to apps log??

hvanyo (approved):

examples/test-site/src/data/site/disabled-pages.json
Guilherme-Almeida-Zeni: @hvanyo ,

Actually, I got this from Allen's previous PR about redirects: https://github.com/johnsonandjohnson/Bodiless-JS/pull/1293/files. I understand this file is not related to redirects itself but I preferred to keep his change since he might have added this file for some reason. If you think it's better, I can remove it right away from this PR.

packages/bodiless-components/src/RedirectAlias/form.tsx
Guilherme-Almeida-Zeni: @hvanyo ,

I really don't know, but I don't see many cases in which this catch would be triggered, I only kept it from the original implementation, so console.error might be enough. Actually, in the original implementation, there was not even a console, it was just returning false. Please, let me know if you want more discussion on this.

examples/test-site/src/data/site/disabled-pages.json
hvanyo: ok leave the json file with the empty content

packages/bodiless-components/src/RedirectAlias/form.tsx
hvanyo: no, i think its fine to leave if we don't think it will ever happen -- and it will never show on static site where we dont' want extraneous console log errrors.  I was just asking the right spot if the purpose was for addressing an issue -- I would expect it to be apps.log and not console.

examples/test-site/src/data/site/disabled-pages.json
Guilherme-Almeida-Zeni: Hello @hvanyo ,

Done. Please, review it again.

hvanyo (approved):

hvanyo (approved):

TatsianaMiraniuk (approved):

Issue #1299, opened 2021-12-14
Filters missing on test site products page


dewen:
## Description

Filters are not displayed on products static page, while shown on edit site.

### Steps to reproduce

- Checkout `main` branch and run setup/start
   `npm run setup && npm run start`
- Open edit site products page: http://localhost:8005/products
- Add filters (tags)
- Build static site and serve: 
   `npm run build`
- Open product page on **desktop**: http://localhost:9000/products/

### Expected result

Added filters displayed on products page

### Actual result

Filters missing.

<img width="962" alt="filter-1" src="https://user-images.githubusercontent.com/1812250/146019451-6308ce89-b52b-406d-a242-504d860c7f67.png">
<img width="888" alt="filter-2" src="https://user-images.githubusercontent.com/1812250/146019467-9748654e-dab4-4c5a-951b-d41eb6ac21ab.png">

TatsianaMiraniuk:
this is a duplicate of #1068

PR #1339, opened 2022-01-14
fix(filtering): Fix FilterByGroup body accordion hidden on load

lucas-varela:
## Changes

Fix an issue where FilterByGroup filters appear in an aria hidden accordion by default on desktop, which cause accessibility issues. This issue only happens on production, since it's caused by SSR. 

Note that this change doesn't affect how filter categories are rendered on mobile.

## Test Instructions

- Run the development server with `npm run dev`
- Go to `/filter-page/`
- Enter edit mode and create some categories
- Quit the development server and create a new build with `npm run build`
- Run a production server with `npm run serve`
- Open `/filter-page/` again using a large browser window (at least 1024px wide)

Before the change:
- The categories are hidden and inaccessible via the keyboard

After the change
- The categories appear normally and are accessible using a keyboard


Guilherme-Almeida-Zeni (approved):

hvanyo:
@lucas-varela @TatsianaMiraniuk  linked https://github.com/johnsonandjohnson/Bodiless-JS/issues/1299

TatsianaMiraniuk (approved):

PR #1358, opened 2022-01-30
feat(Layout, Content library): content Library item created on a deleted page uses src of this …

dewen:
…(deleted) page


packages/bodiless-backend/src/backend.js
```diff
@@ -780,6 +782,22 @@ class Backend {
     });
   }
 
+  static copyAssets(route) {
+    route.post(async (req, res) => {
+      const {path_from, path_to} = req.body;
```
Guilherme-Almeida-Zeni, line 787: Hello @dewen ,

You could deconstruct body first to avoid unexpected issues:
```
const { body } = req;
const { path_from, path_to } = body;
```

BTW, the variables shouldn't be in camel case, like pathFrom and pathTo?

packages/bodiless-backend/src/backend.js
```diff
@@ -780,6 +782,22 @@ class Backend {
...
+      const assetStaticPathFrom = path.join(backendStaticPath, path_from);
+      const assetStaticPathTo = path.join(backendStaticPath, path_to);
+      logger.log(`Copy asset from: ${assetStaticPathFrom} to ${assetStaticPathTo}, cwd: ${process.cwd()}`);
```
Guilherme-Almeida-Zeni, line 790: Hi @dewen ,

This is minor, but I think the text might be "Copy assets..." instead of "Copy asset...".

packages/bodiless-backend/src/backend.js
```diff
@@ -780,6 +782,22 @@ class Backend {
...
+      try {
+        await copyFile(assetStaticPathFrom, assetStaticPathTo);
+        res.send();
+      } catch (error) {
+        logger.log(error);
+        res.status(500).send(`${error}`);
```
Guilherme-Almeida-Zeni, line 812: Hello @dewen ,

Using literals here is really necessary? I think that `.send(error);` should suffice.

packages/bodiless-layouts/src/ContentLibrary/withContentLibrary.tsx
```diff
@@ -32,36 +33,12 @@ export type ContentLibraryOptions = {
...
-  ));
-};
-
-const copyNode = (source: ContentNode<any>, dest: ContentNode<any>, copyChildren: boolean) => {
-  dest.setData(source.data);
-  if (copyChildren) {
-    childKeys(source).forEach(key => copyNode(source.child(key), dest.child(key), true));
-  }
-};
+// const copyNode = (source: ContentNode<any>, dest: ContentNode<any>, copyChildren: boolean) => {
```
Guilherme-Almeida-Zeni, line 36: Hi @dewen ,

Since you have moved copyNode to NodeTools file, why keeping it commented out here?

packages/bodiless-layouts/src/ContentLibrary/withContentLibrary.tsx
```diff
@@ -142,6 +119,6 @@ const withContentLibrary = (options: ContentLibraryOptions) => <P extends object
 
 export default withContentLibrary;
 
-export {
-  copyNode,
-};
+// export {
```
Guilherme-Almeida-Zeni, line 122: Hi @dewen ,

Same as previous comment, since you have moved copyNode to NodeTools file, why keeping the export commented out here?

packages/bodiless-layouts/src/utils/BackendClient.ts
```diff
@@ -46,4 +46,18 @@ export default class BackendClient {
...
+
+  /**
+   * Copy asset.
+   *
+   * @param pathFrom string
+   * @param pathTo string
+   * @returns Promise<AxiosResponse<any>>
+   */
+  copyAsset(pathFrom: string, pathTo: string) {
+    return this.post(`${this.prefix}/assets/copy/`, {
```
Guilherme-Almeida-Zeni, line 58: @dewen ,

If you update the backend.js variables to pathFrom and pathTo, as suggested in previous comment, you just need to pass them directly, like: { pathFrom, pathTo }`

packages/bodiless-layouts/src/utils/NodeTools.ts
```diff
@@ -0,0 +1,93 @@
...
+import BackendClient from './BackendClient';
+
+/**
+ * Check if the node is an asset with image info.
+ * @todo: check 'src' property presents?
+ *
+ * @param node ContentNode
+ * @returns boolean - true if it is an image node.
+ */
+const isAssetNode = (node: ContentNode<any>): boolean => {
```
Guilherme-Almeida-Zeni, line 25: Hello @dewen ,

This function name doesn't seem to reflect what this function does... your function checking if the node is an image, but assets can be anything... I would recommend to update the function name to something like `isImageNode`

packages/bodiless-layouts/src/utils/NodeTools.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param copyChildren boolean - true if child to be moved as well.
+ */
+export const moveNode = async (
+  source: ContentNode<any>,
+  dest: ContentNode<any>,
+) => {
+  try {
+    if (
+      isAssetNode(source)
+      && (source.data.src.indexOf(source.baseResourcePath) !== -1)
```
Guilherme-Almeida-Zeni: Hello @dewen ,

Shouldn't this check be part of isAssetNode function? Note you have added a todo there to implement this...
Also, if you move this to isAssetNode, remember to deconstruct src, like:
```
const { data } = source;
const { src } = data;
```

packages/bodiless-layouts/src/utils/NodeTools.ts
```diff
@@ -0,0 +1,93 @@
...
+    ) {
+      const destDataSrc = source.data.src.replace(source.baseResourcePath, 'site/');
```
Guilherme-Almeida-Zeni: Hello @dewen ,

Could you please deconstruct src?

packages/bodiless-layouts/src/ContentLibrary/withContentLibrary.tsx
dewen: good catch, updated

packages/bodiless-layouts/src/utils/NodeTools.ts
dewen: That makes sense, thanks!

packages/bodiless-layouts/src/utils/NodeTools.ts
dewen: `(source.data.src.indexOf(source.baseResourcePath) !== -1)` is to check if image exists on asset page path, I have removed it from here.

packages/bodiless-layouts/src/utils/BackendClient.ts
dewen: @Guilherme-Almeida-Zeni as these names are for query string parameters, I find some suggestions regarding avoid camel case naming. Maybe we can discuss on dev forum.

Guilherme-Almeida-Zeni (approved):

packages/bodiless-layouts/src/utils/BackendClient.ts
Guilherme-Almeida-Zeni: @dewen , got it. I approved the PR for now, but I think it would be better to discuss this with the dev team and update it later, if needed.

packages/bodiless-layouts/src/utils/NodeTools.ts
```diff
@@ -0,0 +1,177 @@
...
+ * Check if the node is an asset with image info.
+ * @todo: check 'src' property presents?
+ *
+ * @param node ContentNode
+ * @returns boolean - true if it is an image node.
+ */
+export const isImageNode = (node: ContentNode<any>): boolean => {
+  const { data: { src = '' } } = node;
+  const ext = extname(src).toLowerCase().replace('.', '');
+  return !!(src && (imageExtensions.indexOf(ext) !== -1));
```
Guilherme-Almeida-Zeni, line 34: Hello @dewen ,

This double !! seems really weird, but it exists on javascript [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_NOT#double_not_!!). It seems you are trying to enforce return to be boolean but, since the inner condition already returns boolean from && comparison, is this !! really needed?

Guilherme-Almeida-Zeni (dismissed):

Guilherme-Almeida-Zeni:
BTW, 

@dewen , jobs run have failed. Could you please double-check it?

Guilherme-Almeida-Zeni (dismissed):

Guilherme-Almeida-Zeni (approved):

irkhodzka (approved):

PR #1397, opened 2022-02-17
chore(page-data-provider): Refactor Page Data Provider

Guilherme-Almeida-Zeni:
## Overview
This PR contains page data abstraction from gatsby. Additionaly, we also provide refactoring to remove redirect alias form (and related operations) to bodiless page package, fix some unit tests, and fix an issue on "add or create" input field regarding styles.

## Changes
**For the data abstraction, similar solution has been cloned from git provider:**
- Though, only the following fields were abstracted: template; and subtemplate;
- Removed the workaround to have fixed default template on new page form;
- Related page data provider types have been provided. To avoid confusion, existing PageData has been named to PageProps.

**For redirect aliases form refactoring, we needed to:**
- Move all RedirectAlias directory out from bodiless component and place it inside bodiless page source;
-  Update imports in all exported members from RedirectAlias;
-  Export Tools from the bodiless component package so we could import DefaultNormalHref and withToolsButton on page package;
- Get rid of flow from loadsh, since we don't need this package dependency on page package. And to do that, update flow to flowHoc (also in ComponentFormDefaultPanelSize token on bodiless ui);
- Except for the changes mentioned above, nothing else was changed to avoid unexpected behavior.

**For unit tests:**
- Commented out tests to check unnecessary renders and added a todo comment to further investigate the triggering of extra render;
- Update placeholder prop to "placeholderText" on react tags test of props.

**For "add or create" field issue:**
- The wrapper class of the input field has been changed as part of another fix. The older class was "bl-react-tags__search-input " but it now belongs to the input and the wrapper got the class "react-tags__search-wrapper", so it was only needed to update it in the styles to fix the issue.

## Test Instructions
- Retest template on new page operations;
- Retest redirect alias form (and compare to main);
- Retest issue related to "add or create" field.

## Related Issues
N/A


packages/gatsby-theme-bodiless/src/dist/GatsbyPageProvider.tsx
```diff
@@ -13,14 +13,12 @@
  */
 
 import React, { FC, useContext } from 'react';
-import type { GitContextProps } from '@bodiless/core';
```
wodenx, line 16: Why do we still need the GatsbyPageProvider?  I would get rid of it and just use thePageDataProvider, and add a pagePath (which is the same as the slug) to it.

wodenx (changes requested):

lucas-varela (approved):

packages/bodiless-components/src/index.ts
```diff
@@ -121,7 +121,7 @@ export * from './List';
 export * from './Link';
 export * from './FileUpload';
 export * from './PageDisable';
-export * from './RedirectAlias';
```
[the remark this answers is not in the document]
Guilherme-Almeida-Zeni: @wodenx ,

The RedirectAlias directory has been moved to page package, so in this PR it doesn't exist anymore in bodiless component.

packages/gatsby-theme-bodiless/src/dist/GatsbyPageProvider.tsx
Guilherme-Almeida-Zeni: @wodenx ,

We actually don't. I removed it. Please, check it again. Thanks!

wodenx (approved):

PR #1402, opened 2022-02-24
feat(@bodiless/component, @bodiless/filtering): refactor responsive util

dewen:


packages/bodiless-filtering/src/FilterByGroup/token.tsx
```diff
@@ -84,23 +81,13 @@ const asAccessibleFilterByGroup = flowHoc(
...
-    }),
-  ),
-  ifViewportIs(['lg', 'xl', '2xl'])(
-    withDesign({
-      FilterBody: replaceWith(Span),
-      RefineButton: replaceWith(() => null),
-    }),
-  ),
-);
+const asResponsiveFilterByGroup = withDesign({
```
wodenx, line 84: @dewen  I don't understand how this is responsive anymore.  Don't we still need two versions, probably with wrappers which can take responsive classes.

wodenx: or maybe we could just use the existing withResponsiveVariants for that.  In any event - i would maybe move the responsiveness out of @bodiless/filtering for now and put it into canvax and or site/level.

dewen: updated, I applied with withResponsiveVariants in filter token and removed the responsive hoc from FilterByGroupClean.

also asResponsiveFilter is removed since it's not used anywhere.

packages/bodiless-filtering/src/FilterByGroup/token.tsx
```diff
@@ -124,8 +106,30 @@ export const withSingleAllowedTag = flowHoc(
   }),
 );
 
+const asDesktopFilterByGroup = withDesign({
```
wodenx, line 109: We don't need
```
FilterBody: replaceWith(Span)
```
?

packages/bodiless-filtering/src/FilterByGroup/token.tsx
```diff
@@ -124,8 +106,30 @@ export const withSingleAllowedTag = flowHoc(
...
+
+const asMobileFilterByGroup = withDesign({
+  FilterWrapper: asAccordionWrapper,
+  FilterTitle: asResponsiveAccordionTitle,
+  FilterBody: asExpandedOnDesktopBody,
+  ResetButton: asExpandedOnDesktopResetButtonBody,
+  RefineButton: addPropsIf(() => true)(useRefineButtonProps),
+});
+
+const asResponsiveFilterByGroup = (breakpoints: BreakpointsType) => flowHoc(
```
wodenx, line 121: Probably eventually we should remove all of these to cx level.

wodenx (approved):

wodenx (approved):

wodenx (approved):

Issue #1372, opened 2022-02-07
BodilessJS Remove AsBodilessComponent Edit code from Static bundle


CourtLC77:
User Story

AS A Developer
I WANT the AsBodilessComponent Edit code to be executed only when the condition process.env.NODE_ENV !== 'production' is meet.
SO THAT Webpack can remove unused code through the plugin DefinePlugin
Background/Description
Static site includes the code used only on edit environment. It increase the bundle size for unused code.
make the edit code conditional to process.env.NODE_ENV !== 'production' allow Terser to remove unused code.

Acceptance Criteria
AC1. The AsBodilessComponent Edit code is not included in the generated JS Bundle
AC2. The images and links render and behave properly on both static and edit sites.
AC3 Gatsby links are removed from test site.
   
Implementation Notes
POC: https://github.com/marcopagliarulo/Bodiless-JS/commit/69cbc174ca7c02ee76fb8fbc60c0a31a242b2b6f

This should use the static replacement mechanism created in scope of #1440.  This was disabled on the test site pending remediation of gatsby links - it shojuld be re-enabled in scope of this ticket, and gatsby links removed.

Also - let's i mprove the excludsion mechanism to allow both "include" and "exclude" options.

**Testing Notes**
Test Images and Links that they render properly on the static site; this can be a spot check not a full regression;
These sub-defects of #1440  should be verified
- slate editor should render properly in Card Submenus
- slate editor should render properly in Product Cards

PR #1440, opened 2022-03-09
feat: Add static build mechanism

lucas-varela:


packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
```diff
@@ -31,26 +33,42 @@ export const isStaticClientSide = !!(
...
-);
+const withoutHydrationServerSide: WithoutHydrationFunction = (
+  { WrapperElement = DEFAULT_WRAPPER } = {}
+) => WrappedComponent => props => {
+  const { path } = useNode().node;
+  const { nodeKey = '' } = {...props};
+  const id = `${path.join('-')}-${nodeKey}`;
+
+  return (
+    <WrapperElement data-no-hydrate id={id}>
```
wodenx, line 44: @marcopagliarulo should we use a different attribute here or even just mke this the value of the data-no-hydrate ttribute?  i'm a bit worried about using id.

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
```diff
@@ -31,26 +33,42 @@ export const isStaticClientSide = !!(
...
 } = {}) => <P,>(WrappedComponent: ComponentOrTag<P>) => {
   const WithoutHydration: FC<P & WithoutHydrationProps> = (props) => {
     const { forceHydration = false } = props;
-    const rootRef = useRef<HTMLElement>(null);
+    const rootRef = useRef<HTMLDivElement&HTMLSpanElement>(null);
     const [shouldHydrate, setShouldHydrate] = useState<boolean | undefined>(undefined);
 
+    const { path } = useNode().node;
+    const { nodeKey = ''} = {...props};
+    const id = `${path.join('-')}-${nodeKey}`;
```
wodenx, line 62: It would probably be better if the logic that sets the id were encpsulated rather than repeated

packages/bodiless-hydration/src/withoutHydration/types.ts
```diff
@@ -16,7 +16,8 @@ import { HOC } from '@bodiless/fclasses';
 
 type WithoutHydrationOptions = {
   onUpdate?: (props: Record<string, any>, element: HTMLElement | null) => void
-  disableFallback?: boolean
+  disableFallback?: boolean,
+  WrapperElement?: 'div'|'span',
```
wodenx, line 20: @marcopagliarulo  and @lucas-varela  are we happy with limiting this to div or span, or should any element be allowed?

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
```diff
@@ -61,27 +79,80 @@ const withoutHydrationClientSide: WithoutHydrationFunction = ({
       onUpdate(props, rootRef.current);
     });
 
+    useLayoutEffect(() => {
```
wodenx, line 82: Would be great to have a comment here explaining what we are doing...

packages/bodiless-webpack/src/addStaticReplacementPlugin.ts
```diff
@@ -19,33 +19,35 @@ import path from 'path';
...
 const REPLACEMENT = '.static';
 
-type ExcludeSetting = string[] | RegExp;
+type IncludeSetting = string[] | RegExp | boolean;
 
 type PluginOptions = {
   enabled?: boolean
   logging?: boolean
-  exclude?: ExcludeSetting;
+  include?: IncludeSetting;
```
wodenx, line 27: I think a more common use-case is to exclude just a single packge, so I'd prefer we keep this as an `exclude` parameter -- or, better, allow specifying both.

wodenx (dismissed):

packages/bodiless-webpack/src/addStaticReplacementPlugin.ts
wodenx: let's skip this for now - i've added a note to https://github.com/johnsonandjohnson/Bodiless-JS/issues/1372

packages/bodiless-hydration/src/withoutHydration/types.ts
marcopagliarulo: I think it's fine. It will prevent issues with dangerouslySetInnerHTML that doesn't allow to set inner html containing unallowed tags, e.g. set p tag withing a p wrapper.

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
marcopagliarulo: this is required because then the element is retrieved using document.getElementById. Using the attribute data-no-hydrate means that we should traverse the whole tree to retrieve the element. What's your concern about using id?

wodenx: my concern is that id's must be unique, and we are using one here without concern for what other code might do to the dom.  Couldn't we just use document.querySelector?  Or are you concerned that will be significantly less performant?  Anyway, I guess this is not a critical concern.

marcopagliarulo: actually querySelector is faster than getElementById, I am fine to use data attribute to store the Id and then querySelector

wodenx (approved):

wodenx (approved):

irkhodzka (approved):

PR #1488, opened 2022-03-25
test(hydration): Add withoutHydration unit tests

lucas-varela:
## Changes

- Unit tests using withoutHydration were added. 
- The `disableFallback` option was removed from withoutHydration because it isn't useful on Bodiless, since it would only disable rendering components on the Edit environment.
  - No components were using this option, so this change won't affect anything.
- The `config/jestTestSetup.js` file was updated, so it only mocks the `getSelection` implementation on `jsdom` environments. This is required because one of the new tests runs on a node environment.


packages/bodiless-hydration/package.json
```diff
@@ -28,6 +28,11 @@
     "@bodiless/fclasses": "^1.0.0-beta.1",
     "@bodiless/core": "^1.0.0-beta.1"
   },
+  "devDependencies": {
+    "cheerio": "1.0.0-rc.3",
+    "enzyme": "^3.9.0",
+    "react-dom": "^17.0.2"
```
dewen, line 34: Can we move this to peerDeps?

wodenx (approved):

PR #1497, opened 2022-03-30
test(webpack): Add unit tests for the StaticReplacementPlugin

lucas-varela:
## Changes

- Adds unit tests for the StaticReplacementPlugin
- Fix StaticReplacementPlugin only warning about missing `.static` files when `logging` is explicitly `true`.
  - Before the fix, warnings would also be skipped if no `logging` option was passed to the plugin. Now, the only way to skip warnings is to pass `logging: false`.


hvanyo:
@lucas-varela is this still relevant?

Issue #1430, opened 2022-03-07
CanvasX Heros : IMage & Video 


hvanyo:
## Description
As site builder I want hero component to start building my site so that I can create my site more efficiently.

## Acceptance criteria
1. CXstarter has a Homepage uses the Generic Template which has full-bleed Hero.

3. CX-image package has HeroImage matching designs.
     - Image, Landscape as gatsby image
   
4. Cx-youtube package has cxVideo matching designs.
    - with Youtube video: 
    - 16x9 video 
    - standard default player settings include Closed Captioning
      -  cc_load_policy: 1, controls: 1, loop: 0, enablejsapi: 1, modestbranding: 1, rel: 0, origin: getOrigin(),
    - with Full screen enablement
   - with Youtube Video Schema
5. cx-structuredata package exists in bodiless 

6. Cx Hero variations (Hero Image & Video) are in canvasx-starter package, and a styleguide page for each hero component, and available on CanvasX site generic template in the top container slot via chameleon.

7. Documentation on Homepage template & Hero Components.
 
## Design:
- See JIRA ticket

## Hero variations: 
- Hero Image Only
   - Technically this just a landscape image but make a variatian call hero Image.
   - There is potential that variation might offer potential Art Direction Images. 
- Hero Video Only
  - Technically this just a video with all the composed tokens setting the defaults.

PR #1500, opened 2022-03-30
feat(vital-hero, vital-image, vital-card): Add Hero & Card component and update Image 

lucas-varela:
## Changes

This PR adds 2 packages:
- @bodiless/vital-structuredata
- @bodiless/vital-youtube

It also adds Hero tokens to both `@bodiless/vital-image` and `@bodiless/vital-youtube`. Examples on how to use these tokens were added to their respective package docs.

Visit `/styleguide/hero/` for working examples of Image and Video heros. The homepage of the Vital website also contains a chameleon which lets you choose between the two hero variations.

Notes:
1. Structured data for Video heroes is only fetched if you set `BODILESS_GOOGLE_YOUTUBE_API_KEY` with a valid API key into the site's `.env` file.
2. To be more in line with how other styleguide example pages work, instead of creating a separate styleguide page for each Hero variation like AC5 asks, only one was created, which contains all Hero variations.

See https://github.com/johnsonandjohnson/Bodiless-JS/issues/1430 for more information.

## Related Issues
Closes https://github.com/johnsonandjohnson/Bodiless-JS/issues/1430


packages/vital-hero/package.json
```diff
@@ -0,0 +1,41 @@
...
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/johnsonandjohnson/bodiless-js.git"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "build": "tsc -p ./tsconfig.json && npm run build:api-doc",
```
wodenx, line 24: let's follow scripts pattern from here: https://github.com/johnsonandjohnson/Bodiless-JS/blob/chore/upgrade-to-gatsby-4-and-tw-3/packages/vital-flowcontainer/package.json#L24-L30

packages/vital-hero/src/components/Hero/HeroClean.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { asVitalTokenSpec } from '@bodiless/vital-elements';
+import { HeroComponents, HeroProps } from './types';
+
+const heroComponents: HeroComponents = {
+  Wrapper: Section,
+  Content: Fragment,
+};
+
+const HeroCleanBase: FC<HeroProps> = ({ components: C }) => (
+  <C.Wrapper>
```
wodenx, line 29: I don't think we need the Hero component at all.  There's already a wrapper on the slot in the template,  Instead, let's create `vitalImage.Hero` and `vitalYouTube.Hero` tokens and apply these directly to a chameleon on the `TopContent` slot in the generic template, using `on`.

packages/vital-hero/src/components/Hero/tokens/vitalHero.ts
```diff
@@ -0,0 +1,63 @@
...
+  Spacing: {
+    Wrapper: 'pb-9'
+  }
+});
+
+const Default = asHeroToken({
+  ...Base,
+});
+
+const Image = asHeroToken({
```
wodenx, line 33: so this becomes `vitalImage.Hero` -- but we might not even need that - is it any different from `vitalImage.Landscape"?

wodenx (changes requested):

packages/vital-hero/src/components/Hero/HeroClean.tsx
lucas-varela: The `vital-hero` package was created because we needed to place `HeroClean` and `asHeroToken` somewhere, and placing them inside `vital-elements`, for instance, would create a circular dependency between that package and `vital-image`.

Anyway, I also believe we don't need the whole Hero component just for a wrapper since `GenericTemplate` already provides one, so I removed the `vital-hero` package completely and moved the tokens to the image and youtube packages, like you suggested.

packages/vital-hero/src/components/Hero/tokens/vitalHero.ts
lucas-varela: Yeah, it's pretty much a combination of the `Default` and `WithLandscapePlaceholder` image tokens, but I believe creating a pattern where token collections can export a `Hero` token makes it easy to understand where that token might be used. 

That being said, I've moved the Image Hero token to the `vitalImage` token collection after removing the hero package, see: https://github.com/johnsonandjohnson/Bodiless-JS/pull/1500/files#diff-d327313a148c1e0a4a8e140a36565c2baf68ce60b13d486311bf6a99e8da3d51R191

packages/__vital__/src/components/Hero/tokens/__vital__Hero.ts
```diff
@@ -0,0 +1,8 @@
+import { vitalImage } from '@bodiless/vital-image';
```
wodenx, line 1: Seems we no longer need this __vital__ Hero component, right?

packages/__vital__/src/styleguide/Page/StyleGuideTemplate/Hero.tsx
```diff
@@ -0,0 +1,43 @@
...
+  Img,
+} from '@bodiless/fclasses';
+import { asStyleGuideTemplateToken, vitalStyleGuideTemplate } from '@bodiless/vital-templates';
+import { vitalImage } from '@bodiless/vital-image';
+import { YouTubeClean, vitalYouTube } from '@bodiless/vital-youtube';
+
+const ImageHero = as(vitalImage.Hero)(Img);
+const VideoHero = as(vitalYouTube.Hero)(YouTubeClean);
+
+const HeroVariations = (props: any) => (
```
wodenx, line 29: I think maybe instead of having a separate "Hero" page in the styleguide, we should just add these variatios to the Img and YouTube pages.

packages/vital-youtube/src/components/YouTube/tokens/vitalYouTube.ts
```diff
@@ -0,0 +1,109 @@
...
+  withFullScreenEnabled as withBodilessFullScreenEnabled,
+  withYouTubePlayerSettings,
+} from '@bodiless/youtube';
+import { asYouTubeToken } from '../YouTubeClean';
+import { defaultPlayerSettings } from '../util';
+
+/**
+ * WithDefaultPlayerSettings provides standard default YouTube player settings.
+ */
+const WithDefaultPlayerSettings = asYouTubeToken({
```
wodenx, line 35: if these are really the default settings, why aren't we including them in the Default token?  And do we really need a separate token for this?

packages/vital-youtube/src/components/YouTube/tokens/vitalYouTube.ts
```diff
@@ -0,0 +1,109 @@
...
+    Wrapper: WithVideoSchema,
+  },
+  Meta: flowHoc.meta.term('SEO')('With Schema'),
+});
+
+/**
+ * Token that provides VitalDS YouTube base component.
+ */
+const Base = asYouTubeToken({
+  Components: {
```
wodenx, line 67: This should probably be Core

packages/vital-youtube/src/components/YouTube/tokens/vitalYouTube.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Responsive16By9Embed provides responsive YouTube component in widescreen format (16:9).
+ */
+const Responsive16By9Embed = asYouTubeToken({
+  ...Base,
+  Components: {
+    _: as(
+      Base.Components?._,
+      asBodilessResponsive16By9Embed,
```
wodenx, line 93: I'd put this into 'Compose'

packages/__vital__/src/components/Hero/tokens/__vital__Hero.ts
lucas-varela: Including a Hero component in the starter package is part of AC4. @hvanyo how should we proceed here?

packages/__vital__/src/styleguide/Page/StyleGuideTemplate/Hero.tsx
lucas-varela: This is also part of AC5. @hvanyo any thoughts?

packages/vital-youtube/src/components/YouTube/tokens/vitalYouTube.ts
lucas-varela: cc-ing @Guilherme-Almeida-Zeni since he was in charge of the YouTube package.

packages/vital-youtube/src/components/FlowContainer/index.ts
```diff
@@ -0,0 +1,82 @@
...
+const screenVariations = {
+  Default: vitalYouTube.Default,
+  Responsive16By9Embed: vitalYouTube.Responsive16By9Embed,
+};
+
+/**
+ * Flow Container variations for YouTube components.
+ */
+const WithYouTubeVariations = asFluidToken({
+  Components: varyDesigns(
```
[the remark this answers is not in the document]
lucas-varela: cc-ing @Guilherme-Almeida-Zeni since he was in charge of the YouTube package.

packages/vital-youtube/src/components/YouTube/tokens/vitalYouTube.ts
lucas-varela: cc-ing @Guilherme-Almeida-Zeni since he was in charge of the YouTube package.

packages/vital-youtube/src/components/YouTube/tokens/vitalYouTube.ts
lucas-varela: cc-ing @Guilherme-Almeida-Zeni since he was in charge of the YouTube package.

packages/__vital__/src/components/Hero/tokens/__vital__Hero.ts
hvanyo: I don't see any harm having this cmponent available to select from in starter.  we except vital hero card to come in here as well.   There is potentially this hero image/youtube is optimized for full width and has some features that the generic image/youtube don't have.

packages/__vital__/src/styleguide/Page/StyleGuideTemplate/Hero.tsx
hvanyo: Main goal is activation/demonstrable/test.. 

@lucas-varela  and i did write 'pages' but left it to him to decide how.  

I guess keeping images / videos separated is ok.   How about move ImageHero to Images page (which already exists) and  rename this page Video

packages/vital-youtube/src/components/FlowContainer/index.ts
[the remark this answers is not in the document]
hvanyo: I think we only need to provide one player:
DefaultSettings  (default settings & full screen)
SEO schema
Reponsive16x9

If sitebuilder wants something else -- they can build their own variation.  in 95% of cases content editor should choose the above ^

packages/vital-youtube/src/components/FlowContainer/index.ts
```diff
@@ -12,71 +12,22 @@
...
 const WithYouTubeVariations = asFluidToken({
-  Components: varyDesigns(
-    baseVariation,
-    settingsVariations,
-    seoVariations,
-    screenVariations,
-  ),
+  Components: {
+    YouTube: on(YouTubeClean)(
+      vitalYouTube.Default,
```
wodenx, line 25: Le's compose all these tokens into default and just apply Default here. (so in 'vitalYouTube.tsx' add these to the 'Default' token).

packages/vital-structuredata/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@bodiless/vital-structuredata",
```
wodenx, line 2: Consider renaming this @bodiless/structure-data -- I'm not sure there is anything 'vital' specific here, and it could be used with other design systems.  cc @lucas-varela  @hvanyo

packages/vital-youtube/src/components/YouTube/tokens/vitalYouTube.ts
```diff
@@ -0,0 +1,96 @@
...
+});
+
+/**
+ * Token that provides VitalDS YouTube default component.
+ */
+const Default = asYouTubeToken({
+  ...Base,
+  Components: {
+    Item: withYouTubePlayerSettings(defaultPlayerSettings),
+  },
```
wodenx, line 52: Add compose key with all the tokens which are currently added when creating the flow container variation

packages/vital-structuredata/package.json
wodenx: Or maybe even @bodiless/schema-org

wodenx (changes requested):

wodenx (dismissed):

Guilherme-Almeida-Zeni (approved):

packages/vital-image/src/components/Img/tokens/vitalImage.ts
```diff
@@ -161,6 +165,25 @@ const WithLink = asElementToken({
...
+ * component. After refreshing the page, it is rendered as a GatsbyImage, which is full-width by
+ * default. Although not desirable, this behavior is expected.
+ *
+ * Applying this token makes this <img> tag full-width, so it renders just like a GatsbyImage. This
+ * is only required if you don't want the image to "change its size" when refreshing the page
+ * after placing it.
+ */
+const WithFullWidthImage = asElementToken({
+  Layout: {
+    _: withDesign({
```
wodenx, line 181: Hm...  where is this 'Image' design key defined?  Is it part of 'asGatsbyImage? '  If so, then I think we should consider defining asImageToken which allows you to apply these withoout using 'withDesign'.  It's confusing that we are applying a design in an element token...

packages/vital-image/src/components/Img/tokens/vitalImage.ts
```diff
@@ -63,6 +64,9 @@ const Base = asElementToken({
       withGatsbyImageLogger(),
     ),
   },
+  Schema: {
```
wodenx, line 67: Need to check whether this conflicts with work @marcopagliarulo  is doing - see https://github.com/johnsonandjohnson/Bodiless-JS/pull/1581/files#diff-b6a631087f3162c62a6696bed4f400f461b910442b6e5877b81367bd0cc51e41R22-R27.  Originally, @marcopagliarulo  had added this node key to the base token and i asked him to move it - but I believe I was mistaken -- it seems we need this node key pretty much anywhere the image is used.

Guilherme-Almeida-Zeni (approved):

dewen (approved):

solutionsbyheidi (approved):

dewen (approved):

hvanyo (approved):

PR #1515, opened 2022-04-05
feat(gatsby-theme-bodiless): Replace crypto-browserify with crypto-js

marcopagliarulo:
## Changes
crypto-browserify bring inside the bundle a dependencies to two versions of bn.js and readable-stream, resulting in bn.js included in the bundle 6 times, readable-stream 2 times.

Instead of using crypto and then fallback it using crypto-browserify, use directly crypto-js solve the issue.

E.g. test-site bundle is reduced by 737.25 KB to 577.27 KB

### DIFF
--- a/packages/bodiless-backend/src/fileHelper.js
+++ b/packages/bodiless-backend/src/fileHelper.js
@@ -15,7 +15,7 @@
 const fs = require('fs');
 const fse = require('fs-extra');
 const path = require('path');
-const crypto = require('crypto');
+const MD5 = require('crypto-js/md5');
 
 const backendStaticPath = process.env.BODILESS_BACKEND_STATIC_PATH || '';
 
@@ -27,7 +27,7 @@ const copyFilePromise = (from, to) => new Promise((resolve, reject) => {
   });
 });
 
-const generateHash = str => crypto.createHash('md5').update(str).digest('hex');
+const generateHash = str => MD5(str).toString();
 
 const isImage = fileType => {
   const imageFileTypes = [
--- a/packages/bodiless-backend/src/fileHelper.js
+++ b/packages/bodiless-backend/src/fileHelper.js
@@ -15,7 +15,7 @@
 const fs = require('fs');
 const fse = require('fs-extra');
 const path = require('path');
-const MD5 = require('crypto-js/md5');
+const crypto = require('crypto-js');
 
 const backendStaticPath = process.env.BODILESS_BACKEND_STATIC_PATH || '';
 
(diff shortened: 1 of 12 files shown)


packages/bodiless-backend/src/fileHelper.js
```diff
@@ -27,7 +27,7 @@ const copyFilePromise = (from, to) => new Promise((resolve, reject) => {
   });
 });
 
-const generateHash = str => crypto.createHash('md5').update(str).digest('hex');
+const generateHash = str => MD5(str).toString();
```
dewen, line 30: do we need toString here?

packages/gatsby-theme-bodiless/src/dist/DefaultContent/createPlugins.ts
```diff
@@ -13,9 +13,9 @@
  */
 
 import path from 'path';
-import crypto from 'crypto';
+import MD5 from 'crypto-js/md5';
 
-const hash = (name: string) => crypto.createHash('md5').update(name).digest('hex');
+const hash = (name: string) => MD5(name).toString();
```
dewen, line 18: do we need toString call here?

packages/bodiless-layouts/src/deserializers/createFlowContainerItem.ts
```diff
@@ -30,9 +30,8 @@ export type CreateFlowContainerItem = (args: CreateFlowContainerItemArgs) => Flo
 export const generateUuid = (
   content: string,
   index: number,
-) => createHash('md5')
-  .update(JSON.stringify({ content, index }))
-  .digest('hex');
+) => MD5(JSON.stringify({ content, index }))
```
dewen, line 33: shall we add toString here if MD5 returns WordArray object

marcopagliarulo: it is at line 34

packages/bodiless-backend/src/fileHelper.js
marcopagliarulo: Considering that toString performance is fine I'd keep it to avoid possible future regression in case of usage of generateHash in other context.

packages/gatsby-theme-bodiless/src/dist/DefaultContent/createPlugins.ts
marcopagliarulo: Considering that toString performance is fine I'd keep it to avoid possible future regression in case of usage of generateHash in other context.

packages/bodiless-backend/src/fileHelper.js
```diff
@@ -15,7 +15,7 @@
 const fs = require('fs');
 const fse = require('fs-extra');
 const path = require('path');
-const crypto = require('crypto');
+const MD5 = require('crypto-js/md5');
```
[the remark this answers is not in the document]
marcopagliarulo: fixed, thanks

dewen (approved):

Issue #1476, opened 2022-03-21
There is a warning in a console when switching from Edit mode to Review Mode ('Cannot update a component (`WithBreadcrumbItemsFromStore`) while rendering...')


TatsianaMiraniuk:
## Description
There is a warning in a console when switching from Edit mode to Review Mode (any page)

### Steps to reproduce
1. open Test Site Edit env on any page,  e.g. /flow-container/ page
2. open a console in a web-browser
3. click Edit button to switch to Edit mode
4. click Edit button back to switch to Review mode

### Expected result
no warnings

### Actual result
there is a warning 
'Warning: Cannot update a component (`WithBreadcrumbItemsFromStore`) while rendering a different component (`Stylable`). To locate the bad setState() call inside `Stylable`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render'
![Screen Shot 2022-03-21 at 5 09 41 PM](https://user-images.githubusercontent.com/3224040/159364541-caf394a3-48fb-4de1-850e-aaf19169b68e.png)

### Affected Version
reproduced in Main

PR #1529, opened 2022-04-06
fix(navigation): Fix breadcrumb throwing warning on edit

lucas-varela:
Fix https://github.com/johnsonandjohnson/Bodiless-JS/issues/1476 by removing a duplicate observer.


packages/bodiless-navigation/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -160,7 +159,6 @@ const withoutFinalTrail = addProps({
  * HOC that adds breadcrumb props retrieved from breadcrumb store.
  */
 const asBreadcrumbs = flowRight(
-  observer,
```
wodenx, line 163: @lucas-varela  Where is this duplicated?  I don['t see it in withBreadcrumbItemFromStore...

PR #1540, opened 2022-04-09
feat(vital-meta): Add vital-meta package

wodenx:
## Changes
- Add vital-meta package
- Activate in Helmet


packages/vital-meta/src/components/MetaHelmet/helpers.tsx
```diff
@@ -0,0 +1,95 @@
...
+  UploadArea: () => <DropZoneUploadArea>Drag a file or click here to upload.</DropZoneUploadArea>,
+};
+
+export const SocialShareFormImage = () => {
+  const { ComponentFormText } = useMenuOptionUI();
+  const fieldApi = useFieldApi(metaSocialShareImageName);
+  return (
+    <DropZoneWrapper>
+      <ComponentFormText field={metaSocialShareImageName} id="social-share-img-src" />
+      <ImageDropZone
```
hvanyo, line 67: @wodenx   There is open gap that that uses relative url if you upload the image (https://jira.jnj.com/browse/AESQ-3845)    I suggest we remove the ability to upload image and they need to put in absolute image url (maybe with help text "Provide an absolute image url).  Alternatively we fix gap and provide absolute url but don't think its work it.    FYI -- on your branch the image field isn't saving in edit mode.

packages/vital-meta/src/components/MetaHelmet/tokens/vitalMetaHelmet.ts
```diff
@@ -0,0 +1,184 @@
...
+  Compose: {
+    WithOrganizationAreaServed,
+    WithOrganizationContactOption,
+    WithOrganizationContactType,
+    WithOrganizationTelephone,
+  },
+  Flow: flowIf(useIsHomePage),
+});
+
+const WithCountry = asSimpleToken(withMetaStatic({ name: 'bl-country' })({ nodeKey: 'country', nodeCollection: 'site' }));
```
hvanyo, line 63: Purge withcountry & withbrand? -- feel very use case specific.

hvanyo (approved):

PR #1543, opened 2022-04-11
feat(vital-list): Basic vital list component and tokens

wodenx:


packages/vital-list/tailwind.config.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable import/no-dynamic-require, global-require */
```
Guilherme-Almeida-Zeni, line 1: Hello @wodenx ,

Do we need this eslint disable here?

packages/vital-list/tailwind.config.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Copyright © 2021 Johnson & Johnson
```
Guilherme-Almeida-Zeni, line 3: Hello @wodenx ,

Should the year be 2022?

packages/vital-list/src/components/TitleInfographic/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './index.bl-edit';
```
Guilherme-Almeida-Zeni, line 1: @wodenx ,

Instead of all modules, wouldn't be better to specify specific ones to export?

packages/vital-list/getDocs.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable import/no-dynamic-require, global-require */
```
Guilherme-Almeida-Zeni, line 1: @wodenx ,

Do we need this eslint disable here?

Guilherme-Almeida-Zeni (commented):
@wodenx ,

In addition to the review, there are several files missing header license (`npm run copyright` and commit the files).

Also, you are adding a new package but I don't any updates to the root tsconfig.settings.json, I think you should add vital-list to the existing group of paths there.

packages/vital-list/tailwind.config.js
wodenx: i think so yes

packages/vital-list/getDocs.js
wodenx: yes

Issue #1510, opened 2022-04-04
Add selectors to containers on flow container page


pomi:
Description
Add selectors or any other specific attribute to flow container elements (containers)

Steps to reproduce
Open Flow Containers page
Switch to edit mode
Check containers elements

Expected result
There is no distinctive way to select any present container

Actual result
aria-label=ContainerName would be great

Environment
System:
OS: macOS 12.2.1
CPU: (10) arm64 Apple M1 Pro
Shell: 5.8 - /bin/zsh
Binaries:
Node: 16.13.2 - /usr/local/bin/node
npm: 8.1.2 - /usr/local/bin/npm
Languages:
Python: 2.7.18 - /usr/bin/python
Browsers:
Chrome: 99.0.4844.74
Safari: 15.3

Affected Version

commit 22384f184e73366bc00ec619e7b08fa3e5dfc29d (origin/main, origin/HEAD, main)
Author: Lucas Varela <lvarelab@its.jnj.com>
Date:   Mon Mar 21 13:30:11 2022 -0300

    feat: Add static replacement mechanism (#1440)
    
    * Add a mechanism to prevent component hydration

HalinaL:
Closed as completed

PR #1544, opened 2022-04-11
fix: add container selectors on flow-container page

konst8:
## Changes
Add container selectors on the flow-container page.

## Related Issues
https://github.com/johnsonandjohnson/Bodiless-JS/issues/1510


hvanyo (commented):
@konst8  Can we try put the identifiers into flowcontainer component if possible?

dewen (approved):

konst8:
> @konst8 Can we try put the identifiers into flowcontainer component if possible?

@hvanyo in my understanding, in order to do that we will have to add the "...rest" of props here:
https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/bodiless-layouts/src/FlowContainer/EditFlowContainer.tsx#L95
which may possibly cause some regression

sites/test-site/src/data/pages/flow-container/index.tsx
```diff
@@ -112,84 +112,115 @@ const FlowContainerPage = (props: any) => (
...
-        getDefaultWidth={() => 'w-full lg:w-3/4'}
-      />
-      <H2>Contentful Flow Container</H2>
-      <ContentfulFlowContainer />
-      <H2>Limited Flow Container</H2>
-      <FlowContainerLimited nodeKey="limited" />
-      <H2>Nested Flow Container with Default Items</H2>
-      <RegionContainer nodeKey="region" />
+
+      <div aria-label="default-fc-wrapper">
```
hvanyo, line 116: Let's just switch over id's for now.   This isn't a good case for using accessilbiity to identifier containers.

konst8: Updated. Made the flowContainer allowing id prop, removed the wrapper with aria-label. 
Regarding the styleguide/flowcontainer page of __cxstarter__ site - it uses generic template and flow containers are added through the editor interface there, so it won't be easy to add the same ids there. As far as I can see.

hvanyo (approved):

packages/bodiless-layouts/src/FlowContainer/StaticFlowContainer.tsx
```diff
@@ -31,7 +31,7 @@ const flowContainerComponentStart: FlowContainerComponents = {
 
 const NodeProvider = withNode(React.Fragment);
 
-const StaticFlowContainer: FC<DesignableProps> = ({ design }) => {
+const StaticFlowContainer: FC<DesignableProps & { id?: string }> = ({ design, id }) => {
```
dewen, line 34: shall we have a default value for `id`? since it's optional by type

konst8: @dewen i think undefined is okay as default - the id will not be added to the element in this case, which is desirable behavior for us.

PR #1552, opened 2022-04-12
Fix/safelist - not working yet

hvanyo:


packages/bodiless-ui/bodiless.tailwind.config.js
```diff
@@ -60,20 +60,19 @@ module.exports = {
...
-    '../bodiless-carousel/lib/**/!(*.d).{ts,js,jsx,tsx}',
-    '../bodiless-core/lib/**/!(*.d).{ts,js,jsx,tsx}',
-    '../bodiless-core-ui/lib/**/!(*.d).{ts,js,jsx,tsx}',
-    '../bodiless-layouts/lib/**/!(*.d).{ts,js,jsx,tsx}',
-    '../bodiless-layouts-ui/lib/**/!(*.d).{ts,js,jsx,tsx}',
-    '../bodiless-richtext/lib/**/!(*.d).{ts,js,jsx,tsx}',
-    '../bodiless-richtext-ui/lib/**/!(*.d).{ts,js,jsx,tsx}',
-  ],
+  // idea from: https://github.com/tailwindlabs/tailwindcss/discussions/6347#discussioncomment-2185719
+  ...(process.env.NODE_ENV === 'development') && {
```
hvanyo, line 64: 64 is not working but if we had another way to detect it on start vs build

the safelist feature with pattern works as described and keeps the bl-classes in edit mode.  Just need to turn off safelist feature for static.

hvanyo:
@wodenx 
with second commit here 

- Edit works and not bl-classes are not purged.
- Static has bl-classes when I search

Issue #1523, opened 2022-04-06
Public Demo Site


wodenx:
## User Story
As a prospective site builder
I want a public site demonstrating key Bodiless features
So that I can learn how amazing Bodiless really is

## ACs
There is a public demo site deployed to p.sh and netlify which showcases the following features:
- Sitewide data in header/footer
- Rich text editing
- Plain text editing
- Gatsby image
- Flow container
- List editing
- Navigation (cx-navigation)
- Content library

Publicly available site should not allow saving content (BODILESS_BACKEND_SAVE_ENABLED="0") -- this requires #1522

kbray0009:
Hey I have created a new card.

https://6o2fc3.axshare.com/#id=y2wl2m&p=fake_brand_site&g=1

kbray0009:
Table design

https://dbzu8p.axshare.com/#id=pobau4&p=page_1&g=1

hvanyo:
Closing as we don't need this issue and site will live at https://github.com/wodenx/bodiless-demo

PR #1603, opened 2022-04-19
chore(bodiless-psh): deploy site on platform.sh based on env variable

konst8:
## Changes
Select which site to deploy based on env variable.

## Related Issues
https://github.com/johnsonandjohnson/Bodiless-JS/issues/1523


dewen (approved):

hvanyo (approved):

edit/force-build
```diff
@@ -0,0 +1 @@
+19_Apr_2022
```
Asemirski, line 1: What is that?

hvanyo: @Asemirski  A date :)     ha ha  this is just a temp file we update to force a build on platform.sh with a commit and not touch real files.  Its been our habit to use a date sometimes I have seen hello...

PR #1621, opened 2022-04-22
feat(i18n): Implement i18n

konst8:
## Changes
Implement i18n.


packages/bodiless-i18n/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "name": "@bodiless/i18n",
+  "version": "1.0.0-beta.9",
+  "description": "Library to provide multilingual support for Bodiless",
+  "author": "Konstantin Mishur <kmishur@its.jnj.com>",
+  "license": "Apache-2.0",
+  "main": "lib/index.js",
+  "typings": "lib/index.d.ts",
+  "files": [
+    "/lib",
+    "/bodiless.docs.json"
```
wodenx, line 11: need to add getDocs

packages/bodiless-i18n/src/LanguageNode/withLanguageNode.tsx
```diff
@@ -0,0 +1,34 @@
...
+} from '@bodiless/core';
+import { HOC } from '@bodiless/fclasses';
+import { LanguageContentNode } from './LanguageContentNode';
+import { useLanguageContext } from '../LanguageProvider';
+
+type NodeProps = {
+  nodeKey?: string,
+  nodeCollection?: string,
+};
+
```
wodenx, line 15: Please add jsdoc blocks for all exports, including usage examples where appicable.

packages/vital-layout/src/components/LanguageSelector/LanguageSelectorClean.tsx
```diff
@@ -0,0 +1,62 @@
...
+const languageSelectorComponents: LanguageSelectorComponents = {
+  Wrapper: Span,
+  Link: A,
+};
+
+/**
+ * @private
+ * Base logo component.
+ */
+const LanguageSelectorBase: FC<LanguageSelectorProps> = ({ components: C }) => (
```
wodenx, line 47: Do we need this component at all if it is just a link with a wrapper?  Perhaps we can simply create a vitalLink token to implement language selection.

packages/vital-layout/src/components/LanguageSelector/tokens/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  Content: {
+    Link: addPropsIf(() => true)(useLanguageLinkProps),
+  },
+  Theme: {
+    Wrapper: 'text-m-base border-l-2 border-vital-primary-divider lg:border-r-2'
+  },
+  Spacing: {
+    Wrapper: 'pl-5 lg:mr-5 lg:px-5 lg:py-2',
+  },
+  Behavior: {
```
wodenx, line 46: I tink this should be in the  "Content" domain,  Also - let's discuss at standup where these shoudl be specified -- maybe at the __vital__ level?

packages/vital-layout/src/components/LanguageSelector/tokens/index.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const Default = asLanguageSelectorToken({
+  Content: {
+    Link: addPropsIf(() => true)(useLanguageLinkProps),
+  },
+  Theme: {
+    Wrapper: 'text-m-base border-l-2 border-vital-primary-divider lg:border-r-2'
+  },
+  Spacing: {
```
wodenx, line 43: If we get rid of the 'LahguageSelector' component, then we could add a LanguageButtonWrapper in HeaderClean and put these classes there.  Again, something to discuss at standup.  cc @hvanyo

packages/vital-layout/src/components/LanguageSelector/tokens/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  return {
+    children: secondLanguage.label,
+    href: secondLanguage.isDefault
+      ? pagePathWithoutPrefix
+      : `${secondLanguage.name}${pagePathWithoutPrefix}`,
+  };
+};
+
+const Default = asLanguageSelectorToken({
+  Content: {
```
wodenx, line 37: I think this should be in Core domain.  Also, I believe you can now just use `addProps`  -- it will now take a function.

packages/vital-layout/src/components/Header/HeaderClean.tsx
```diff
@@ -47,7 +48,7 @@ const headerComponents: HeaderComponents = {
   UtilityMenu: MenuClean,
   DesktopSearch: DesktopSearchClean,
   SearchToggler: SearchTogglerClean,
-  LanguageButton: Fragment,
+  LanguageButton: LanguageSelectorClean,
```
wodenx, line 51: let's rename this key to 'LanguageSelector' (not `LanguageButton`)

sites/__vital__/src/data/pages/es/index.json
```diff
@@ -0,0 +1,3 @@
+{
```
wodenx, line 1: Let's be sure to add this also to vital-demo.

packages/vital-layout/src/components/LanguageSelector/tokens/index.ts
wodenx: I think the right way to do this is at site level (in the __vital__ starter (and also in the vital-demo).  So probbly in the default Header there is no language selector, just a token which can be used to add one, and at site level the header is shadowed to add that token, along with the specific language choices (fine to use es/en for starter -- for demo we need to discuss what 2nd language to support).

wodenx: [7:33 AM] Oden, Christopher [CPCUS]
    so probably the pattern should be:

	wrap whole page in language context
	wrap secitnos of page (hader/footer) in language nodes which allow per-langauge content.

does that make sense?
​[7:34 AM] Mishur, Konstantin [JJCRU Non-J&J]
    yes, absolutely
​[7:34 AM] Oden, Christopher [CPCUS]
    if so then i think we need 3 tokens

	page level token which adds the provider
	header/footer tokens which add the langauge nodes

And i think these should not be applied by default in the vital- packagds -- but applied at site level (through shadowing) on starter and/or demo -- but that is to be discussed at stanup
<https://teams.microsoft.com/l/message/19:a13da064505048bdbb55feb0428ac740@thread.tacv2/1650886421202?tenantId=3ac94b33-9135-4821-9502-eafda6592a35&amp;groupId=e0707489-d17d-44c8-9066-df88931a96a7&amp;parentMessageId=1650884152314&amp;teamName=CanvasX Core Platform Team&amp;channelName=CanvasX Dev&amp;createdTime=1650886421202>

wodenx (changes requested):

packages/bodiless-i18n/package.json
konst8: Added.

packages/bodiless-i18n/src/LanguageNode/withLanguageNode.tsx
konst8: Added. May be needs to be extended in the docs.

packages/vital-layout/src/components/LanguageSelector/LanguageSelectorClean.tsx
konst8: Done.

packages/vital-layout/src/components/LanguageSelector/tokens/index.ts
konst8: Done.

packages/vital-layout/src/components/LanguageSelector/tokens/index.ts
konst8: Done.

packages/vital-layout/src/components/LanguageSelector/tokens/index.ts
konst8: Done.

packages/vital-layout/src/components/Header/HeaderClean.tsx
konst8: Done.

sites/__vital__/src/data/pages/es/index.json
konst8: Done.

packages/__vital__/src/components/Page/tokens/__vital__Page.ts
```diff
@@ -28,13 +28,32 @@ import {
   asGenericTemplateToken
 } from '@bodiless/vital-templates';
 import { useNode } from '@bodiless/core';
+import { withLanguages } from '@bodiless/i18n';
+
+const DefaultTemplate = asGenericTemplateToken({
```
wodenx, line 33: Rather than putting ths in page, it should be in __vital__GenericTemplate.  Also why not sjadow that?

packages/__vital__/src/components/Page/tokens/__vital__Page.ts
```diff
@@ -28,13 +28,32 @@ import {
...
+  ...vitalGenericTemplate.Default,
+  Core: {
+    ...vitalGenericTemplate.Default.Core,
+    PageWrapper: withLanguages([
```
wodenx, line 37: To be safe we should repeat vitalGenericTemplateBase.Default.Core.PageWrapper here or just use extension, as
```
asGenericTemplateToken(vitalGenericTemplateBase.Default, {
  Core: {
    PageWrapper: ...
```

packages/__vital__/src/components/Page/tokens/__vital__Page.ts
```diff
@@ -28,13 +28,32 @@ import {
   asGenericTemplateToken
 } from '@bodiless/vital-templates';
 import { useNode } from '@bodiless/core';
+import { withLanguages } from '@bodiless/i18n';
+
+const DefaultTemplate = asGenericTemplateToken({
+  ...vitalGenericTemplate.Default,
```
wodenx, line 34: Should always extend the ...Base version of the token collection, ie
```
...vitalGenericTemplateBase.Default
```
That version is not shadowed.  Otherwise, if you use this to shadow vitalGenericTemplate, there will be a circular import problem.

packages/__vital__/src/components/Page/tokens/__vital__Page.ts
```diff
@@ -28,13 +28,32 @@ import {
...
+        name: 'es',
+        label: 'Español',
+      },
+    ]),
+  },
+});
 
 // @todo remove NoBreadcrumbsGeneric when breadcrumbs is implemented and
 // content editor can choose to use breadcrumb
 const NoBreadcrumbsGeneric = asGenericTemplateToken({
```
wodenx, line 53: While we're in here - i think we should create a 'vitalGenericTemplate.WithNoBreadcrumbsOnHomePage' token with the condition in the 'Flow' domain, like:
```
const WithNoBreadcrumbsOnHomePage = asGenericTemplateToken({
  Flow: flowIf(isHomePage),
  Components: {
    BreadcrumbWrapper: replaceWith(() => null),
   Breadcrumb: replaceWith(() => null).
  },
});
```
and compose it into the Base token"
```
const Base = asGenericTemplateToken({
  ...,
  Compose: {
    WithNoBreadcrumbsOnHomePage,
   }.
   });
   
    
Compose it into vitalGenericTemplate.Base, and put the condition into the 'Flow' domain. like
const WithNoBrad

File: packages/__vital__/src/shadow/@bodiless/vital-layout/Footer.ts [hunk 13]
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+import { asFluidToken } from '@bodiless/vital-elements';
+import { vitalFooterBase } from '@bodiless/vital-layout';
+import {
+  addProps, as
+} from '@bodiless/fclasses';
+import { withLanguageNode } from '@bodiless/i18n';
+
+const Default = asFluidToken({
```
wodenx, line 21: Shouldn't this be asFooterToken, not asFluid Token?  Also, I think i'd use extension here since we're not overriding anything, just adding -- so
asFooterToken(vitalFooterBase.Default, {
  Core: {
     _: addProps(...),
   },
  ...
});

packages/bodiless-i18n/src/LanguageSelector/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      ? pagePathWithoutPrefix
+      : `${secondLanguage.name}${pagePathWithoutPrefix}`,
+  };
+};
+
+/**
+ * asLanguageSelector is a hoc which creates language toggler
+ * based on LinkClean component.
+ */
+export const asLanguageSelector = on(LinkClean)(
```
wodenx, line 50: This doesn't belong in @bodiless/i18n - LinkClean is a vital component so you are introducing a dependency we don't want.  I think you can probably create a 
/src/components/LanguageSelectorLink in vital-layouts and create a vitalLanguageSelectorLink token collection which imports useLanguageSelectorProps from @bodiless/i18n (or maybe jsut export `asLanguageSelector` from i18n as
```
export const asLanguageSelector = addProps(useLanguageSelectorProps)
```

packages/vital-layout/src/components/Header/HeaderClean.tsx
```diff
@@ -69,7 +69,7 @@ const HeaderCleanBase: FC<HeaderProps> = ({ components: C, ...rest }) => (
           <C.UtilityMenuWrapper>
             <C.UtilityMenu />
           </C.UtilityMenuWrapper>
-          <C.LanguageButton />
+          <C.LanguageSelectorWrapper />
```
wodenx, line 72: Sorr if i misled, i think we should have both `LanguageSelectorWrapper` and `LanguageSelector`:
```
<LangauageSelectorWrapper />
  <LanguageSelector />
</LanguageSelectorWrapper>
```
Not sure if we need to make the wrapper a Div and put the classes on it, or if it's ok to put the classes directy on the LanguageSelector.  Thoughs @hvanyo  ?  In the latter case, the wrapper would jut be a fragment.

packages/vital-layout/src/components/Header/tokens/vitalHeader.ts
```diff
@@ -91,10 +70,11 @@ const Base = asHeaderToken({
     // @todo perhaps this should be an element spcing token ike "LargeIconSize".
     MenuToggler: 'w-6 h-6',
     MenuTogglerWrapper: 'my-4',
+    LanguageSelectorWrapper: 'pl-5 lg:mr-5 lg:px-5 lg:py-2',
   },
   Theme: {
     Wrapper: vitalColor.BgPrimaryPage,
-    LanguageButton: 'lg:border-vital-primary-divider lg:border-r-2',
+    LanguageSelectorWrapper: 'lg:border-vital-primary-divider lg:border-r-2',
```
wodenx, line 77: I would create a composable `WithLanguageSelector` token which can be removed or added at site level.

wodenx (changes requested):
See comments.

packages/__vital__/src/shadow/@bodiless/vital-layout/Header.ts
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { asFluidToken } from '@bodiless/vital-elements';
+import { vitalHeaderBase } from '@bodiless/vital-layout';
+import { asLanguageSelector } from '@bodiless/i18n';
+import { addProps } from '@bodiless/fclasses';
+
+const Default = asFluidToken({
```
wodenx, line 19: Shouldn't this be asHeaderToken?

wodenx: Again, I think you can use extension here.  And probably we can just use the new `WithLanguageSelector` token exported from vitalHeader.

packages/__vital__/src/shadow/@bodiless/vital-layout/Header.ts
```diff
@@ -0,0 +1,54 @@
...
+export const asLanguageSelectorLink = on(LinkClean)(
+  asLinkToken({
+    ...vitalLink.Default,
+    // Make the link not editable.
+    Schema: {},
+  }),
+  asLanguageSelector
+);
+
+const WithLanguageSelector = asHeaderToken({
```
wodenx, line 30: Sorry - I mean't that this token should be exported form vitalHeader, and *appied* at site level...

wodenx (changes requested):
Just move the WithLanguageSelector token into vitalHeader

hvanyo (commented):
Can you shadow __vital__Page instead of doing this:

https://github.com/lucas-varela/Bodiless-JS/pull/29
should closely align with your PR for shadowing..

grab my vital-templates/Page.ts & change in the sites/template/__default file.

packages/__vital__/src/shadow/@bodiless/vital-layout/Footer.ts
```diff
@@ -22,14 +23,11 @@ const Default = asFooterToken(vitalFooterBase.Base, {
...
-    FooterMenu: as(
+    _: as(
       withLanguageNode,
-      vitalFooterBase.Base.Schema.FooterMenu
-    ),
-    Copyright: as(
-      withLanguageNode,
-      vitalFooterBase.Base.Schema.Copyright,
-    ),
+      withNodeKey({ nodeKey: 'Footer', nodeCollection: 'site' }),
```
wodenx, line 28: Usual pattern is to apply the node key in the enclosing component - so in this case it would be in vitalLayout.

wodenx: but leave the `withNode` here.

wodenx: actually i'm a bit confused by this... should't the withNode be inside (*before* when using `as`) the withNodeKey?

packages/__vital__/src/shadow/@bodiless/vital-layout/Header.ts
```diff
@@ -27,25 +28,25 @@ export const asLanguageSelectorLink = on(LinkClean)(
...
-  WithLanguageSelector,
+  vitalHeaderBase.WithLanguageSelector,
+  {
+    Core: {
+      _: addProps({ 'data-shadowed-by': '__vital__Header' }),
+    },
+    Schema: {
+      _: as(
+        withLanguageNode,
+        withNodeKey({ nodeKey: 'Header', nodeCollection: 'site' }),
```
wodenx, line 41: Agin, node key should probably go in vitalLayout - and I"m not sure the withNode is necessary.  Maybe this is taken care of by withLanguageNode...

wodenx (changes requested):

packages/__vital__/src/shadow/@bodiless/vital-layout/Header.ts
konst8: withLanguageNode does nothing if there is no language prefix in the path, so we still need withNode for default case.

packages/vital-layout/src/components/Layout/tokens/vitalLayout.ts
```diff
@@ -58,6 +59,10 @@ const Base = asLayoutToken({
   Content: {
     Header: addProps({ id: LayoutIds.HeaderContent }),
   },
+  Schema: {
+    Header: withNodeKey({ nodeKey: 'Header', nodeCollection: 'site' }),
```
wodenx, line 63:
```suggestion
    Header: withNodeKey({ nodeKey: 'header', nodeCollection: 'site' }),
```

packages/vital-layout/src/components/Layout/tokens/vitalLayout.ts
```diff
@@ -58,6 +59,10 @@ const Base = asLayoutToken({
...
+    Footer: withNodeKey({ nodeKey: 'Footer', nodeCollection: 'site' }),
```
wodenx, line 64:
```suggestion
    Footer: withNodeKey({ nodeKey: 'footer', nodeCollection: 'site' }),
```

wodenx (dismissed):

packages/__vital__/src/components/Page/tokens/__vital__Page.ts
```diff
@@ -11,44 +11,20 @@
  * See the License for the specific language governing permissions and
```
hvanyo, line 11: @konst8  Page was suppose to move shadowing..   I have in my shadow PR.   I don't know what will merge first but ideally we want nothing in components.    If we don't change, it only concern if this gets caught in testing and shadow merges first we undo shadowing.

hvanyo (approved):

hvanyo (approved):

packages/__vital__/src/shadow/@bodiless/vital-layout/Footer.ts
```diff
@@ -0,0 +1,30 @@
+/**
```
hvanyo, line 1: @konst8  Can you move these to vital_test?

hvanyo: oh never mind all these have i18n in ithem..  they should all stay here.

wodenx (approved):

PR #1637, opened 2022-04-25
chore(demo, starter): Ensure that demo and starter are in sync.

wodenx:
## Test Instructions
Light smoke testing on vital-demo.
Create __vital__ starter and light smoke testing there.
Max 1-2h


dewen (approved):

sites/__vital__/bodiless.env.config.js
```diff
@@ -1,27 +1,6 @@
...
-      gitBranchName = currentBranch;
-    } else {
-      console.warn('You are in "detached HEAD" state...');
-    }
-  } catch (e) {
-    console.warn(e);
-  }
-
-  return gitBranchName;
-};
```
marcopagliarulo, line 24: This block of code is removed from __vital__ and added to vital-demo. It would create a diff between the two sites, is that expected?

wodenx: yes -  this is the one intended difference -- the demo site will not allow saving on the main branch.  This is to allow us to deploy a public version which does not permit any kind of persistence, but is probably not somehting we need to activate by default on starter.

marcopagliarulo (approved):

Issue #1608, opened 2022-04-19
Contentful Card: the app crashes if Homepage with a Contentful Card that got a new image is being cloned.


TatsianaMiraniuk:
**Steps to reproduce:**
1. open Home page (with Contentful card) in Edit Env Edit Mode Test Site 
2. upload a new image into Contentful Card
3. in Page, click Clone
4. add a new page path
5. Save
 
**Actual Result:**
the app crashes with an error
![Screen Shot 2022-04-19 at 2 45 20 PM](https://user-images.githubusercontent.com/3224040/164074203-6e12e0d6-3358-4ce7-ac98-85c9dbd1f951.png)

**Expected Results:**
the page clones without errors

NOTE: 
'npm run start' fails after restarting.

PR #1641, opened 2022-04-26
fix(bodiless-backend): Fix image assets paths to be replaced on clone operation

Guilherme-Almeida-Zeni:
## Overview
There is an issue to clone the homepage in case there is an image uploaded to any components. This is happening because replacing function is not updating properly the page. We are refactoring the backend function to update the cloned assets to use simple template string instead of regex.

## Changes
- Replace regex with template strings to build the paths to replace/be replaced;
- Additionaly, we update bodiless page client type to include clone

## Test Instructions
N/A

## Related Issues
- #1608


dewen (approved):

Guilherme-Almeida-Zeni:
> Fix to main branch?

@dewen , updated to upgrade branch.

konst8 (approved):

PR #1645, opened 2022-04-26
fix(bodiless-psh): hide git credential on platform.sh

dewen:


dewen:
changes applied from https://github.com/johnsonandjohnson/Bodiless-JS/pull/1632/files

PR #1671, opened 2022-04-29
chore: Create gatsby starter.

### LINKED ISSUE #1643: Create Minimal Gatsby Starter
As a developer
I want to be able to create a bodiess site from a standard Gatsby starter
So it's easy as pie to get started.

AC1.
There exists a gatsby-starter-bodiless-minimal repository which can be cused to create a site using the minimal starter via `gatsby new`,

AC2.
The above site can be hosted easily on gatsby cloud.

AC3.
The base repository of the above site can be easily updated by running bodiless new and pushing the resuts.

wodenx:
## Changes
- Update minimal starter so it can deploy to gatsby cloud

## Test Instructions
- Clone repo and checkout and setup this branch
- From bodiless-cli directory
  ```
  npm run bl new -- --clone-local -r HEAD --setup "npm run setup:gatsby-cloud" --name minimal
  ```
- Push the new site to a github repository in a separate (non-jnj) account --not to the bodiless repo
- Sign up for free tier gatsby cloud account
- Create a new gatsby cloud project linked to the github repo -- set gatsby root dir to /sites/minimal.
- Verify that new site deploys correctly to gatsby cloud
- Aso verify that site builds and starts locally.

## Related Issues
Closes #1643


hvanyo (approved):

package.json
```diff
@@ -72,12 +74,6 @@
     "symlink-dir": "^3.0.3"
   },
   "devDependencies": {
-    "@babel/core": "^7.9.0",
-    "@babel/plugin-proposal-class-properties": "^7.2.1",
-    "@babel/plugin-proposal-decorators": "^7.4.0",
-    "@babel/plugin-transform-modules-commonjs": "^7.5.0",
```
marcopagliarulo, line 78: why removing the babel plugins? they are defined in babel config
https://github.com/johnsonandjohnson/Bodiless-JS/blob/1ecc5bbdc74d6bfeab60696517b402148a5a6286/babel.config.js

marcopagliarulo (commented):
Please resolve the conflict on packages/gatsby-theme-bodiless/package.json

package.json
wodenx: i moved them to site level.

marcopagliarulo: not all the plugin are moved to the site. 

1. @babel/core is missing from __minimal__
2. @babel/plugin-transform-modules-commonjs is missing at all, it is required by test because we compile to ES, but test needs commonJS, without this plugin, all the tests will fail
3. @babel/plugin-proposal-class-properties is missing as dependency, we got it because it is a dependency of @babel/preset-env that is required by babel-preset-gatsby. I think would be better to list the dependency.

wodenx: - babel-core is required by gatsby, so i'm not sure we need to list it
- i guess we can restore babel-plugin-transform-modules-commonjs
- why do we need to list this as a dependency when we are not using it ourselves?

wodenx: btw i don't think that babel.conifg.js is used by anything -we set the babel config in gatsby-node...

marcopagliarulo: - We don't need to list babel-core, but I think is better to listen the direct dependencies. In this case we need @babel/core at the root of monorepo because we use @babel/plugin-transform-modules-commonjs.
- We can remove from the dependencies @babel/plugin-proposal-class-properties because it is used directly by babel-preset-gatsby, but at this point I don't think we need it as a site dependency as well.
- babel.conifg.js is used when running test, not sure on site build. Anyway, gatsby documentation says to use it in case of a monorepo https://www.gatsbyjs.com/docs/how-to/custom-configuration/babel/.

I think we should verify the exact usage of  babel.conifg.js.

marcopagliarulo (dismissed):

hvanyo (approved):

marcopagliarulo (approved):

PR #1674, opened 2022-05-02
fix(bodiless-backend): copy root repo credential helper to tmp clone

dewen:

### DIFF
--- a/packages/bodiless-backend/src/git.js
+++ b/packages/bodiless-backend/src/git.js
@@ -331,11 +331,11 @@ const mergeMain = async () => {
   // Get root repo credential.helper.
   const credentialHelperCmd = GitCmd.cmd().add('config', '--get', 'credential.helper');
   const credentialHelper = getGitCmdOutputString(await credentialHelperCmd.exec());
-  logger.log(`credentialHelper: ${credentialHelper}`);
   await clone(rootDir, { directory: tmpDir, branch: upstreamBranch.replace('origin/', '') });
   process.chdir(tmpDir);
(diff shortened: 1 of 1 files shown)


packages/bodiless-backend/src/git.js
```diff
@@ -328,10 +328,18 @@ const mergeMain = async () => {
   const remote = getGitCmdOutputString(await GitCmd.cmd().add('remote', 'get-url', 'origin').exec());
   const rootCmd = GitCmd.cmd().add('rev-parse', '--show-toplevel');
   const rootDir = getGitCmdOutputString(await rootCmd.exec());
-
+  // Get root repo credential.helper.
+  const credentialHelperCmd = GitCmd.cmd().add('config', '--get', 'credential.helper');
+  const credentialHelper = getGitCmdOutputString(await credentialHelperCmd.exec());
+  logger.log(`credentialHelper: ${credentialHelper}`);
```
wodenx, line 334: THis doesn't display the credentials does it?

wodenx (approved):

packages/bodiless-backend/src/git.js
dewen: it displays helper value, i.e.
`@sites/test-site: [BACKEND] 2022-05-02T15:42:12.684Z credentialHelper: osxkeychain`

dewen: but in case of credential store with a file, it does show the path to credential, so I made a minor adjustment without showing the details

PR #1685, opened 2022-05-05
fix(gatsby-theme-bodiless): add autoprefixer

hvanyo:
* Add autoprefix to postcss css


packages/gatsby-theme-bodiless/build-css.js
```diff
@@ -40,6 +40,8 @@ if (!tailwindThemeEnabled) {
         postCssPlugins: [
```
wodenx, line 40: I'm assuming this is only added if BODILESS_TAILWIND_THEME_ENABLED is true, yes?
If not, we should probably fix that while we're in here.

wodenx (approved):

PR #1774, opened 2022-05-10
ci: add playwright to action

Asemirski:


dewen (dismissed):

Asemirski:
Please note that playwright requires both "@playwright/test" package and npx playwright install --with-deps to install required runtime

TatsianaMiraniuk (approved):

dewen (approved):

Issue #1564, opened 2022-04-13
"Unhandled Runtime Error" when clicking 'Add' in FlowContainer's Local Context Menu on the linked text (Edit mode) #1564


DavtyanH:
**Steps to reproduce**

1. Run starter site on Edit env and go to relative URL "/styleguide/flow-container/" on Edit mode
2. Add the link to the text on "Content Registration FlowContainer"
3. Click on the linked text 
4. Click on the 'Add' icon in the "Content block
![tempsnip2](https://user-images.githubusercontent.com/38629857/163209228-e3ec98b2-f302-4602-8bb8-02c45c49ddfb.png)
"
**Expected result**
Component Picker open and No errors are presented on the console.

**Actual result**
Shown 'Unhandled Runtime Error'.
![Capture1](https://user-images.githubusercontent.com/38629857/163210012-d0a83a86-7def-46eb-8258-03c84dd75b5b.PNG)

Next error is presented:
`framework.js:5431 Warning: Cannot update a component (`DevOverlay`) while rendering a different component (`Stylable`). To locate the bad setState() call inside `Stylable`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
    at WithGatsbyImageLogger (http://localhost:8005/component---src-templates-styleguide-jsx.js:100557:77)
    at div
    at http://localhost:8005/component---src-templates-styleguide-jsx.js:59660:16
    at Stylable
    at Stylable
    at WithDesign (http://localhost:8005/component---src-templates-styleguide-jsx.js:99788:35)
    at Stylable
    at Stylable
    at Stylable
    at Stylable
    at WithDesign (http://localhost:8005/component---src-templates-styleguide-jsx.js:99788:35)
    at Stylable
    at Stylable
    at WithoutProps
    at WithData (http://localhost:8005/component---src-templates-styleguide-jsx.js:57686:30)
    at Trigger (http://localhost:8005/component---src-templates-styleguide-jsx.js:134626:92)
    at Tooltip (http://localhost:8005/component---src-templates-styleguide-jsx.js:133741:32)
    at wrappedComponent (http://localhost:8005/component---src-templates-styleguide-jsx.js:119237:73)
    at WithLocalContextMenu
    at wrappedComponent (http://localhost:8005/component---src-templates-styleguide-jsx.js:119237:73)
    at WithMeta
    at wrappedComponent (http://localhost:8005/component---src-templates-styleguide-jsx.js:119237:73)
[...]

PR #1776, opened 2022-05-10
fix(bodiless-core): Unhandled Runtime Error when node data is not available

marcopagliarulo:
## Changes
Set data as an empty object when retrieved if not available. Handle on gatsbyImage and BaseList the node.data as optional.

## Related Issues
#1564


packages/bodiless-core/src/ContentNode.ts
```diff
@@ -120,7 +120,7 @@ export class DefaultContentNode<D extends object> implements ContentNode<D> {
 
   get data() {
     const { getNode } = this.getters;
-    return getNode(this.path) as D;
+    return getNode(this.path) || {} as D;
```
wodenx, line 123: Is this necessary since we already have [this](https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/bodiless-core/src/Store/BodilessMobxStore.ts#L130) ?

wodenx (approved):

packages/bodiless-core/src/ContentNode.ts
marcopagliarulo: that one is the default getter, but the getter can be custom on a different instance of DefaultContentNode.
In the specific case, since the nodeProvider gets the current collection if not specified, and the selected component is the richtext, opening the component selector, the current is the richtext one, that uses a custom getter [here](https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/bodiless-richtext/src/RichTextItemGetters.bl-edit.tsx#L60)

This solution prevents issue whatever the getter is

TatsianaMiraniuk (approved):

Issue #1542, opened 2022-04-09
BodilessJS Remove AsBodilessComponent Edit code from Static bundle


wodenx:
## Description
https://github.com/johnsonandjohnson/Bodiless-JS/issues/1372

 
## Acceptance Criteria
AC1. The AsBodilessComponent Edit code is not included in the generated JS Bundle
 AC2. The styles of text in rich text are the same on static and edit.h3. Implementation Notes

PR #1785, opened 2022-05-16
feat(bodiless-core): BodilessJS Remove AsBodilessComponent Edit code from Static bundle #1542

marcopagliarulo:
## Changes
AsBodilessComponent has a static version
Static replacement is enabled on test site
Richtext has a static version (required to enable static replacement on test site)
Static Replacement allow the option exclude to exclude specific files from static replacement
WithoutHydration is refactored

## Related Issues
#1542


packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
```diff
@@ -80,14 +87,9 @@ const fullSelector = (element: HTMLElement | null) => {
...
 
-const useWrapperId = (element: HTMLElement | null) => {
-  const selector = fullSelector(element);
-  return createHash('md5').update(selector || '').digest('hex');
-};
-
-const withoutHydrationServerSide: WithoutHydrationFunction = (
-  { WrapperElement = DEFAULT_WRAPPER } = {}
-) => WrappedComponent => props => (
+const withoutHydrationServerSide: WithoutHydrationFunction = ({
```
wodenx, line 90: It's a bit confusing that a function called `withoutHydrationServierSide` is used on the client side in edit mode.  Can we come up with a better name?

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
```diff
@@ -80,14 +87,9 @@ const fullSelector = (element: HTMLElement | null) => {
...
+  WrapperElement = DEFAULT_WRAPPER,
```
wodenx, line 91: I think we should move the default options resolution to `withoutHydration` rather than repeating it in both `...serverSide` and `...clientSide`

packages/bodiless-core/src/asBodilessComponent/asBodilessComponent.static.tsx
```diff
@@ -0,0 +1,62 @@
...
+/**
+* Makes a component "Bodiless" by connecting it to the Bodiless-jS data flow and giving it
+* a form which can be used to edit its props. Returns a standard `asBodiless...` function,
+* which takes `nodeKey` and `defaultData` parameters, and returns an HOC which yields an editable
+* version of the base component.
+*
+* @param options An object describing how this component should be made editable.
+*/
+// eslint-disable-next-line max-len
+const asBodilessComponent = <P extends object, D extends object>(options: BodilessOptions<P, D>): AsBodiless<P, D> => (
```
wodenx, line 32: What do you think of the idea of also exporting this version directly -- maybe `asBodilessReadOnlyComponent`?  Maybe not worth it, since there is now no liability in terms of bundle size to just using `asBodilessComponent` even for a site which is *never* editable (eg one that uses a CMS).

packages/bodiless-richtext/__tests__/RichTextStatic.tsx
```diff
@@ -0,0 +1,59 @@
+/**
```
wodenx, line 1: Should this not be named `RichTextStatic.test.tsx"?

wodenx (commented):
See questions.

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
marcopagliarulo: I created another function called withoutHydrationClientSideEdit to handle this case.

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
```diff
@@ -80,14 +87,9 @@ const fullSelector = (element: HTMLElement | null) => {
...
+} = {}) => WrappedComponent => props => (
   <WrapperElement data-no-hydrate style={isEditClientSide ? {} : { display: 'contents' }}>
```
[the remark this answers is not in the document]
marcopagliarulo: yes it is true, it can create potential layout discrepancies but we need to remove { display: 'contents' } otherwise the local context menu would not show up in edit mode. As solution there is a new function returning just the given component without wrapper only on client edit. Potentially { display: 'contents' } could create layout issue to a given component, for example cases with child positioned absolute. To handle this case I added the param WrapperStyle allowing to override the style { display: 'contents' } in order to reset with a custom solution the wrapper style.

packages/bodiless-core/src/asBodilessComponent/asBodilessComponent.static.tsx
marcopagliarulo: I think this is a good idea, especially because it is effortless. The static version of asBodilessComponent now is just a reexport of asBodilessReadOnlyComponent which contain the static logic, so it can be used as asBodilessReadOnlyComponent and as static version at the same time.

packages/bodiless-richtext/__tests__/RichTextStatic.tsx
marcopagliarulo: Fixed, thanks

wodenx (approved):

wodenx (approved):

wodenx (approved):

PR #1817, opened 2022-05-27
feat(fclasses): Overload asTokenSpec.

wodenx:
## Changes
- Allow asTokenSpec to take className strings or HOC's which are applied to _ key of Core domain
- Provide a single 'Core' domain as a default when asTokenSpec is invoked with no arguments.


packages/fclasses/__tests__/tokenSpec.test.tsx
```diff
@@ -87,6 +87,48 @@ describe('extendMeta', () => {
...
+    });
+    expect(test.Core).toBeDefined();
+    // Check that core domain and default domains ar allowed
+    const test1 = asTokenSpec()({
+      Core: {
+        _: 'test',
+      },
+      Flow: flowHoc,
+      Compose: {},
+    });
```
dewen, line 104: Do we expect test1.Core to be 'test' here?

packages/fclasses/__tests__/tokenSpec.test.tsx
```diff
@@ -87,6 +87,48 @@ describe('extendMeta', () => {
...
 
 describe('asTokenSpec', () => {
+  it('Supplies core domain by default', () => {
+    const test = asTokenSpec()({
+      Meta: {
+        title: 'Test',
+      },
+    });
+    expect(test.Core).toBeDefined();
+    // Check that core domain and default domains ar allowed
```
dewen, line 97: ar => are

packages/fclasses/src/tokenSpec.ts
```diff
@@ -249,7 +288,11 @@ const asTokenSpec = <
     spec.Compose = {};
     spec.Flow = undefined;
     spec.Meta = {};
-    mergeWith(spec, ...specs, tokenMergeCustomizer);
+    const coreDomain = Object.keys(d)[0];
```
dewen, line 291: Are we sure the first key is always core domain? I thought user could provide different keyed domains

wodenx: It will use whatever the first key is.  There is no guarantee that there is a 'Core' domain at all.

dewen (dismissed):

packages/fclasses/src/tokenSpec.ts
```diff
@@ -222,24 +222,63 @@ const on = (
...
+ * If specified, defines a circumscribed set of allowed token domains.  If omitted, a single
+ * 'Core' domain will be allowed.  The resulting function will produce a token specification
+ * in which all domains are guaranteed to be present (those which were not defined will be
+ * empty objects). The order of the domains is consistent and defined by the order of the keys
+ * in this parameter.
  *
  * @returns
- * Function which recieves an object as a parameter and returns a normalized token
- * specification based on that object.
+ * Function which receives a set of partial token speciications and returnsa normalized token
```
marcopagliarulo, line 243: Please add a space between returns and a. Fix the typo in specifications

marcopagliarulo (dismissed):

packages/fclasses/src/tokenSpec.ts
wodenx, line 243:
```suggestion
 * Function which receives a set of partial token speciications and returns a normalized token
```

marcopagliarulo (approved):

dewen (approved):

PR #1818, opened 2022-05-27
fix(@bodiless/navigation): warning in a console when switching from E…

dewen:
…dit mode to Review Mode


packages/bodiless-navigation/src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -170,7 +169,6 @@ const withoutFinalTrail = addProps({
  * HOC that adds breadcrumb props retrieved from breadcrumb store.
  */
 const asBreadcrumbs = flowRight(
-  observer,
```
dewen, line 173: observer applied in `[asBreadcrumb](https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/bodiless-navigation/src/Breadcrumbs/asBreadcrumb.tsx#L71)` might cause the warning

dewen: observer has been applied to AsBreadcrumb

https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/bodiless-navigation/src/Breadcrumbs/asBreadcrumb.tsx#L71

dewen:
Fix committed from a different PR

PR #1821, opened 2022-06-02
ci: Manual prerelease

wodenx:


.github/workflows/prerelease.yaml
```diff
@@ -3,6 +3,7 @@ env:
   CI: true
   NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 on:
+  workflow_run
```
Asemirski, line 6: should be workflow_dispatch:

Asemirski (changes requested):

PR #1873, opened 2022-06-07
fix(@bodiless/navigation): warning in a console when switching from Edit mode to Review Mode

dewen:
…dit mode to Review Mode


packages/bodiless-navigation/src/Breadcrumbs/asBreadcrumb.tsx
```diff
@@ -95,7 +94,9 @@ const asBreadcrumb = ({
       store,
     });
 
-    if (!isEdit) {
+    const isProduction = process.env.NODE_ENV === 'production';
+
+    if (isProduction) {
```
wodenx, line 99: What happens if we only do this during SSR?

packages/bodiless-navigation/src/Breadcrumbs/asBreadcrumb.tsx
```diff
@@ -112,7 +113,7 @@ const asBreadcrumb = ({
     useLayoutEffect(() => () => {
       // Only necessary in edit mode since items are not added or removed
       // under any other circumstances.
-      if (isEdit) {
+      if (!isProduction) {
```
wodenx, line 116: I think we should remove this condition entirely -- and I think we can also remove the `isSSR` in the first layout effect, and consolidate to:
```
useLayoutEffect(() => {
  store.setItem(item);
  return () => {
    store.deleteItem(item);
  }
  }, ...)
```

dewen: @wodenx  this change seems bring back the breadcrumb flicker issue on static site.

wodenx (approved):

Issue #1008, opened 2021-06-14
Accordions - Edit Mode: the space cannot be typed in Title(the space key works as a trigger to expand/collapse an Accordion)


TatsianaMiraniuk:
**Steps to reproduce:**
1. in Edit env, Edit Mode,  add an Accordion to a flow container
2. add any text in Accordion Title
3. add a space

**Actual Results:**
the space is not added. Accordion opens (closes) instead.

**Expected Results:**
Accordion Title is editable and allow to type the spaces.

PR #1899, opened 2022-06-08
feat: Add vital-accordion package (copy of PR 1790)

hvanyo:
## Changes

- Adds the `@bodiless/vital-accordion` package, which is standalone and **doesn't** depend on the `@bodiless/accordion` package.
  - It does reuse a fair amount of code from the previous accordion package, but most parts were refactored.
- Adds FAQPage schema to `@bodiless/schema-org`.
- Adds an `/accordion` page in the Vital Demo styleguide.
- Adds accordion variations to the Vital FlowContainer.
- Fixes the pesky issue #1008


dewen:
> ### Test Cases
> ### Verify component picker
> 1. On homepage, in edit mode. Click add on Flowcontainer:
> 2. Verify you have 4 variations – accordion with collapsed/expand on open, faq with collapsed/expanded on open.
> 3. I updated the body to say "Accordion Body" instead of render as RTE toolbar. verify that works in component picker.
> 
> ### Verify accordions are added to flowcontainer
> 1. Add the 4 accordion flowcontainers to new page and confirm variations work as expected. ( done on page /styleguide/accordioninflowcontainer)
> 2. Switched to Preview move and clicked reload and verified correct accordions were expanded/collapsed on open
>    Verify static
> 
> ### Verify static
> 1. Visit http://localhost:9000/styleguide/accordioninflowcontainer and verified correct accordions were expanded/collapsed on open

These cases verified locally

dewen (approved):

PR #1921, opened 2022-06-15
fix(vital-templates): Fix Template Names

hvanyo:
## Changes
- Fixed template names 
- Default -> Generic
- Add meta title to PDP tempalte

## Test Instructions

- In edit mode, on home page, click page | Template &  confirm you see the correct list of template names
- In edit mode, add a new page, click page | Template and choose each template and confirm you get expected templated.


packages/vital-templates/src/components/GenericTemplate/tokens/vitalGenericTemplate.ts
```diff
@@ -120,10 +120,10 @@ const ContentListing = asGenericTemplateToken({
   }
 });
```
hvanyo, line 122: @wodenx  If this was left as "Default" -- the Generic was never picked and always got Base.   Why?

dewen (dismissed):

packages/vital-templates/src/components/GenericTemplate/tokens/vitalGenericTemplate.ts
wodenx: Not sure I follow - what do you mean "Generic was never picked"?

dewen (approved):

PR #1925, opened 2022-06-18
fix(sites): remove set-missing-symlinks from site prepare during lerna bootstrap.

dewen:
## Changes

- @bodiless namespace no longer available for site level symlink during lerna bootstrap. Having this `prepare` blocks `npm install` on site level.

## Test Instructions

1. Setup and build the project
2. Light regression test on test/demo site.
3. Compare the build with main branch build.

Expected
- no tailwindcss config change
- no doc build change
- env .env.* changes.


wodenx:
I question whether we still need 'set-missing-syminks; at all.  I believe reviously, these symlinks were sued for a variety of auto-discovery mechanisms which needed to traverse the node_modules directory, but i believe the logic for all of those things has been revised.

We should test:
- tailwind config generation
- documentation tree generation
- .env file creation
If all of these things work properly, then we can remove set-missing-symlinks entirely.

@marcopagliarulo  Do you agree?

dewen:
> I question whether we still need 'set-missing-syminks; at all. I believe reviously, these symlinks were sued for a variety of auto-discovery mechanisms which needed to traverse the node_modules directory, but i believe the logic for all of those things has been revised.
> 
> We should test:
> 
> * tailwind config generation
> * documentation tree generation
> * .env file creation
>   If all of these things work properly, then we can remove set-missing-symlinks entirely.
> 
> @marcopagliarulo Do you agree?

Removed set-missing-symlinks.js and find no changes to build.

PR #1952, opened 2022-08-08
POC : Vital Decoupled with Card/Buttons/Links [canary]

hvanyo:
## Changes
- Separate out the packages of vital-card, vital-button, vital-link from design & editable.
  - vital-card-core, vital-button-core, vital-link-core now contain the vital design only.
     -  added "asExternalBaseCOMPONENT" token stub that sets data uses withDesign to populate data via addProps.  Could be used as is or eventually replaced/set by vital-component-CMS package. 
  - Existing vital-card, vital-buttons, vital-link add the editing schemas and hydration features and function identical to before (thus no breaking changes)
    - For each package -- export the core collection followed by any tokens that need the editing. (Which includes anything without prefixed "with")  and add editable/flowcontainer tokens as needed.)
- Add sites/gatsby-vanilla site that only includes @bodiless/fclasses, @bodiless/vital-*-core & @bodiless/vital-elements and none of the 'bodiless editable funcitonality'.   
  - Builds with Bodiless tailwind scripts.  
  - Demos vital design with no editing functionality

## Test Instructions
- No Change in Vital Bodiless Demo and functionality
   - npm run start on sites/vital-demo and visit http://localhost:8000/styleguide/card -- confirm no change. 
   - verify npm run build & serve on sites/vital-demo and visit http://localhost:9000/styleguide/card -- confirm no change. 
- Gatsby Vanilla Site demo'ing vital only design + tailwind building from bodiless.
   - npm run start on sites/gatsby-vanilla and visit http://localhost:8000/styleguide/card -- there vital design cards should be displayed with stubbed external data. 
   - verify npm run build & serve on sites/gatsby-vanilla and visit http://localhost:9000/styleguide/card


sites/gatsby-vanilla/src/components/header.jsx
```diff
@@ -0,0 +1,118 @@
...
+    Title: addProps({children: 'Menu item 3'}),
+  }),
+)(MenuTitleClean);
+
+const ExternalMenu = as(
+  vitalMenuCore.TopNav,
+  withDesign({
+    Item: addProps({ test: 'heidiexternalmenu'}),
+    Title: startWith(MenuItem1),
+    // Need to add this Item with Title starting with MenuItem2/MenuItem3
```
hvanyo, line 41: @wodenx  what's a clean way to add more item menus for a static menu?
This works for 1st menu item but how to add more non-editable menus items (such as MenuItem2 & MenuItem3? )

Then this leads me to second question as how to add sublists without editor?
I think if we go down this route, we would want to add API to add static list items & sublists not through editor -- unless we can do that now??

packages/vital-navigation-core/src/components/BurgerMenu/tokens/BurgerMenuContext.tsx
```diff
@@ -0,0 +1,104 @@
+// Copy from bodiless-navigation for the Burger Close transition..
```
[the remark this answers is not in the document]
wodenx: we had this issue before.... i can't remember the cause.  bodiless.tailwind.config.js is only used to generate bodiless admin css.
Actually -  t hink bodiless-navigation depends on bodiless-core (or bodiless-core-ui) which we don't want.  Is there still styling in bodiless-navigation?   I thought we had deprecated it

packages/vital-navigation-core/src/components/Menu/MenuClean.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { asVitalTokenSpec } from '@bodiless/vital-elements';
+import { Nav, Ul, flowHoc } from '@bodiless/fclasses';
+import { asBodilessMenu } from '@bodiless/navigation';
+// import { asStylableList, List } from '@bodiless/components';
+import type { MenuComponents } from './types';
+
+/**
+ * A clean menu for use with all navigation elements.
+ */
+const MenuCleanNonEditable = flowHoc(
```
hvanyo, line 25: @dewen  I was doing some rearranging of withChild/withParent from @bodiless/core -> @bodiless/fclasses (code is not committed) but this line throws the error: 

@bodiless/vital-navigation-core: src/components/Menu/MenuClean.tsx(32,7): error TS2742: The inferred type of 'MenuCleanNonEditable' cannot be named without a reference to '../bodiless-components/node_modules/@bodiless/core/lib'. This is likely not portable. A type annotation is necessary.

Do you know how to fix?

wodenx: What if you try typing this explicitly?  Someting like `: React.CompoenntType<DesignableProps<MenuComponents> & ...>` not sure exactly what the other props are.
But also - `asBodielssMenu()` definitlye depends on @bodiless/core,  so we need to move that too...

PR #1955, opened 2022-09-09
feat(gatsby-theme-bodiless, backend, hydration, layouts, tokens, ga4): Replace crypto-browserify with crypto-js

### LINKED ISSUE #1954: crypto-browserify used on client side as replacement of nodejs crypto introduces outdated and duplicated dependencies
## Description
crypto-browserify has outdated dependencies and cross references with different versions.
Two different versions of bn.js and readable-stream are included in the client bundle. bn.js is included a total time of 6, readable-stream 2 times.
It produces a bigger js bundle that increase the parsing and evaluation time by the browser, badly impacting the performances.

### Steps to reproduce
1. Set env variable BODILESS_BUILD_STATS='1'
2. Set env variable BODILESS_OPEN_STATS='1'
3. Build the site
4. check the automatically onened tab and look for the packages bn.js  and readable-stream

### Affected Version
v1.0.0-rc.15

marcopagliarulo:
## Changes
Crypto-browserify is not under active maintenance (last release 5 years ago). It bring outdated dependencies and multiple version of the same package inside the client bundle.

Replace crypto-browserify with crypto-js, 

### File upload
1. From edit interface pick any component with image
2. Upload an image
3. Check the image is uploaded properly generating the hash value as expected.
### PLP GA4
1. Open PLP with GA4 enabled on static site 
2. Filter the Products
3. Verify that the event pushed contains the item_list_id with the proper hash generated
### withoutHydration
1. Open any page on static site having at least one component withoutHydration
2. verify the ID on the component is the same after load and on the static HTML
### flowContainer Item creation
1. From edit interface add any component to flow container
2. Check the UUID is generated properly inside the JSON files.
### Move files to public
1. Build the site
2. Verify that gatsby has properly copied the files to the public folder.
### DefaultContent
1. Build a site with more than one Default Content directory created by createDefaultContentPlugins
2. Verify that all the defautl content are available
### Tokens
1. On test site, add a component to the tokens page and enable/disable tokens
2. Verify that the component is reloaded properly every time a token is applied/removed.

## Related Issues
Fixes #1954


wodenx (dismissed):

wodenx (approved):

packages/bodiless-backend/src/fileHelper.js
```diff
@@ -15,7 +15,7 @@
 const fs = require('fs');
 const fse = require('fs-extra');
 const path = require('path');
-const crypto = require('crypto');
+const crypto = require('crypto-js');
```
VancheeZze, line 18: Everywhere else we do `import MD5 from 'crypto-js/md5';`. Don't think it impacts tree-shaking in any way though.

VancheeZze (dismissed):

packages/bodiless-backend/src/fileHelper.js
marcopagliarulo: addressed, thanks

VancheeZze (dismissed):

packages/bodiless-backend/__tests__/fileHelper.test.ts
```diff
@@ -0,0 +1,63 @@
+/**
```
wodenx, line 1: Why is this here?

marcopagliarulo: Because fileHelper.js uses MD5, I wrote a test for it which was not covered by any test

wodenx (approved):

VancheeZze (approved):

Issue #1960, opened 2022-09-30
Vital Cards don't work with Content Library as expected


hvanyo:
## Description

A Vital Card when you add to Content Library, the data disappears.

### Steps to reproduce

- Add a vital card to vital flowcontainer (need to do on branch docs/vt-images or had to add ‘vitalFlowContainer.Default’ to the ContentLibraryContainer in packages/vital-test/src/styleguide/Page/StyleGuideTemplate/FlowContainer.tsx.
- Select it and choose Library | Add

### Expected result

Component renders identical to before you added it to library

### Actual result

Component renders with no data  (once you remove it from library then the data reappears)

### Environment
-

### Affected Version

^1.0.0-rc.16

PR #1962, opened 2022-10-17
fix(__vital__): Fix ContentLibrary not working on vital cards

hvanyo:
## Changes
Fix shadowing of cards in __vital__ package so that ContentLibrary works as expected

## Test Instructions
* Verify content library works as expected now without it disappearing as its added to content library.  https://github.com/johnsonandjohnson/Bodiless-JS/issues/1960
* Verify only Basic & Hero type cards are available in vital demo site and they work as expected.


dewen:
Vital page FlowContainer with content library feature tested.

PR #1969, opened 2022-11-09
[WIP] Feature/poc contentful

marcopagliarulo:
On Contentful create a content type 'page' having fields: 

1. **path**, type _Short text_, field option _This field represents the Entry title_, required, Unique field.
2. **template**, type _Short text_, required.

On Contentful create a content type 'node' having fields: 

1. **path**, type _Short text_, field option _This field represents the Entry title_, required, Unique field.
2. **content**, type _JSON Object_, required.

Setup the local environment and build **@sites/bodiless-contentful**.

> Note - after setup, you must rebuild @bodielss/ui in order to properly whitelist bodiless admin css.

On Contentful create a new app from Apps > Manage Apps, click link at up right of the page **Manage app definitions**, then Create app.

APP Settings:
**Hosted by Contentful**: Flagged
**Locations**; Flagged **Page** -> **Show app in main navigation** and **Entry Editor**

Once Hosted by Contentful is flagged, the upload area is enabled, drag the directory public within @sites/bodiless-content-editor inside the drag area.

Save the app.

Install the app into your space.

Now the app should be listed below **Apps** menu Item.

Edit the content type page, select the tab **Entry editors**, Add the new listed app to the ones on the right. Save.

 Run Static site using contentful data:

set environment variables BODILESS_CONTENTFUL_SPACE_ID and BODILESS_CONTENTFUL_ACCESS_TOKEN to connect to contentful


wodenx:
Was able to get this working in our Contentful test account.  It's a great start!  A few comments:
- Changes made in the bodiless editor do not create a new draft of the page.  I assume this is because theydon't modify the page itself, but rather the nodes.
  - What are the advantages/use-cases of storing the individual nodes as a separate content type, rather than serializing the whole store as a single json object to the page?
  - Can we link these nodes to the page via a reference field?  If so, is there a way in contentful that changes to a referenced node can trigger a new revision of the page?
  - The nodes are all created in "published" state, which is obviously not viable as a long-term situation.  How would we envision the workflow in this scenario  -- eg create then as unpubished and then publish all of them (or put all through workflow from the page?  
  - The more I think about it, the more I think saving everything to the page as a single json object is simpler and cleaner and solves a lot of these problems...
- I see the header and footer are editable on the page -- but these are site -level.  Perhaps the best solution here is to link to "header" and "footer" content types which can be edited independentely (show them as read-only here).  Thoughts?
- Is there a way to pop-out the editor into a new browser window so that it can be viewed without the contentful chrome?
- 
More to come - but great work @marcopagliarulo

wodenx:
Some next steps
- replace  image uploads with contentful media assets -- ideally using the contentful asset selector (but using component selector would probably also be ok)
- -

marcopagliarulo:
> Changes made in the bodiless editor do not create a new draft of the page. I assume this is because they don't modify the page itself, but rather the nodes.

Correct

> What are the advantages/use-cases of storing the individual nodes as a separate content type, rather than serializing the whole store as a single json object to the page?

Actually nothing specific, it was just to reflect the json file structure approach. We can save on a single json object with low effort.

> Can we link these nodes to the page via a reference field? If so, is there a way in contentful that changes to a referenced node can trigger a new revision of the page?

We could create a reference field accepting multiple references, but we cannot create new revisions based on references https://www.contentful.com/faq/versioning/#does-entry-versioning-apply-to-linked-content

> The nodes are all created in "published" state, which is obviously not viable as a long-term situation. How would we envision the workflow in this scenario -- eg create then as unpubished and then publish all of them (or put all through workflow from the page?

Well this depends by the data structure that we decide to use, I think if we turn to store as a single json object to the page could simplify the workflow.

> I see the header and footer are editable on the page -- but these are site -level. Perhaps the best solution here is to link to "header" and "footer" content types which can be edited independentely (show them as read-only here). Thoughts?

I would create a contentype site that contains all the header and footer contents.

> Is there a way to pop-out the editor into a new browser window so that it can be viewed without the contentful chrome?

Looks like no because it would not be wrapped into the SDK

PR #1987, opened 2023-02-21
feat(chore): Cleanup Vital Component tokens

VancheeZze:
## Changes
- Validated Domain Definitions.


packages/vital-buttons/src/components/Buttons/tokens/vitalButtons.ts
```diff
@@ -67,15 +67,16 @@ const Base = asButtonToken({
...
   Layout: {
     Wrapper: 'flex-row-reverse',
+    Icon: 'inline-block',
   },
   Components: {
     Icon: replaceWith(Span),
   },
   Theme: {
-    Icon: 'vital-arrow group-hover:text-current text-transparent',
+    Icon: 'w-6 h-2 vital-arrow group-hover:text-current text-transparent',
```
VancheeZze, line 76: I notice we have `width` and `height` in both: `Theme` and `Layout`. The `Theme` seems to appear more often.

packages/__vital__/src/shadow/@bodiless/vital-templates/Page.ts
```diff
@@ -16,6 +16,8 @@ import { withLanguages } from '@bodiless/i18n';
 import { asFluidToken } from '@bodiless/vital-elements';
 import { vitalPageBase } from '@bodiless/vital-templates';
 
+// `vitalPage.Base` is the same as `vitalPageBase.Base`.
```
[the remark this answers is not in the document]
hvanyo: @VancheeZze  confirmed.  This is the Base used for Shadowing..   This is correct.   For shadowing to work the collection had to be named something else or we go into infinite loop.  So we chose vitalPageBase == vitalPage and thus vitalPageBase.Base = vitalPage.Base :).   This is one reason why we are not exporting Base as it means two things.

packages/vital-buttons/src/components/Buttons/tokens/vitalButtons.ts
hvanyo: @VancheeZze  these dictate size of element so that feels more 'themey' than layout.  I could easily change size of something and keep using same layout.   Tailwind defines it as 'sizing' https://tailwindcss.com/docs/width  -- can you update readme of domains to include sizing and then let's follow that pattern.

packages/__vital__/src/shadow/@bodiless/vital-templates/Page.ts
```diff
@@ -16,6 +16,8 @@ import { withLanguages } from '@bodiless/i18n';
...
+// Can we reuse initial default vital components here?
```
hvanyo, line 20: @VancheeZze  Base = Default so it should be fine.
https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/vital-templates/src/components/Page/tokens/vitalPage.ts#L63

packages/vital-card/src/components/Card/tokens/Hero.ts
```diff
@@ -52,6 +52,8 @@ const HeroBase = asCardToken({
     DescriptionWrapper: 'mb-5 lg:mb-6',
   },
   Theme: {
+    // If we edit vitalTypography.H1 and move `vitalFontSize` from 'Core' to 'Theme',
+    // we could do just "TitleWrapper: vitalTypography.H1.Theme," and same for `Meta`
```
hvanyo, line 56: Not following why this is helpful @VancheeZze

packages/vital-elements/src/components/Typography/tokens/vitalTypography.ts
```diff
@@ -38,7 +38,7 @@ const Link = asElementToken({
 
 const H1 = asElementToken({
   Core: {
-    _: vitalFontSize.XXXL,
+    _: vitalFontSize.XXXL, // Should Font Size be in Core for all these tokens?
```
hvanyo, line 41: It feels like it should be theme doesn't it..    This is kind of gray area as these base tokens.    I think I agree it should Theme.  Check  for any omit(H# and see if it impacts usages)

packages/vital-link/src/components/Link/index.ts
```diff
@@ -15,6 +15,8 @@
 export * from './index.bl-edit';
 
 export { default as vitalLink } from './tokens';
+// Some base exports like this are used within a `./vital-test/src/shadow/@bodiless/vital-list/`
+// Should we updadte to use `vitalLink`?
```
hvanyo, line 19: We need to export a copy of collection with 'Base' at end for shadowing (see first comment.  so continue to leave vitalLinkBase as is.

packages/vital-templates/src/components/PDPTemplate/tokens/vitalPDPTemplate.ts
```diff
@@ -53,6 +53,8 @@ const Default = asPDPTemplateToken(vitalGenericTemplate.Base, {
     ProductDetailWrapper: 'lg:pl-2 pt-4 lg:pt-0',
   },
   Theme: {
+    // If we edit vitalTypography.H1 and move `vitalFontSize` from 'Core' to 'Theme',
```
hvanyo, line 56: same question as above

packages/__vital__/src/shadow/@bodiless/vital-templates/Page.ts
VancheeZze: But this is inside the `./shadow/...` directory — would the same rules with naming apply here?

packages/vital-card/src/components/Card/tokens/Hero.ts
VancheeZze: My judgment was that `vitalTypography.H1` includes multiple domains: `Core`, `Theme`, `Spacing`, and `Meta` which are all added to the `Theme` Domain of `HeroBase`.

If we look at the `vitalTypography.H1.Core` — it really only has `_: vitalFontSize.XXXL` which is just a `font-size` that could be moved to the `Theme` domain and if we do so, here, instead of omitting `Spacing` we would do `vitalTypography.H1.Theme`.

What it also does is that now the `Theme` domain of `HeroBase` will only have `Theme` tokens in it, but we will need to add `vitalTypography.H1.Meta` to the `Meta` domain as well.

Just wanted to share some thoughts

packages/vital-buttons/src/components/Buttons/tokens/vitalButtons.ts
VancheeZze: Just to double-check — We are introducing a new domain `Sizing` according to the Tailwind docs, documenting it [here](https://johnsonandjohnson.github.io/Bodiless-JS/#/VitalDesignSystem/Components/VitalElements/Tokens/TokenDomains?id=allowed-domains), and moving all width, height, etc classes into the new domain? 

Could that break things?

hvanyo: @VancheeZze  sorry no thanks for asking -- I reference tailwind doc to see categories to confirm domains if they are question.  (We mapped some of the domains to tailwind css categories.).  Tailwind 'Sizing' is a category we don't have so have to decide if that is Theme or Layout.    I think its more Theme than layout, so move width/height to Theme Domain.  then just update this line https://github.com/johnsonandjohnson/Bodiless-JS/blob/37c76988c6f61937bad520d0521fa34a1775c04f/packages/vital-elements/doc/TokenDomains.md?plain=1#L25-L26 to  say "typography, sizing, etc.."

packages/__vital__/src/shadow/@bodiless/vital-templates/Page.ts
hvanyo: @VancheeZze we should still  do this 
import { vitalPageBase } from '@bodiless/vital-templates';
but this line
const Default = asFluidToken(vitalPageBase.Base, {
should be
const Default = asFluidToken(vitalPageBase.Default, {

hvanyo: collectionNameBase is collection renamed for shadowing.
collectionNameBase.Default should refer to the default token behavior in that collection.

packages/vital-card/src/components/Card/tokens/Hero.ts
hvanyo: I think we solved the issue via chat...

packages/vital-buttons/src/components/Buttons/tokens/vitalButtons.ts
VancheeZze: Done!

packages/vital-card/src/components/Card/tokens/Hero.ts
VancheeZze: Yep, updated as discussed + added some comments.

packages/vital-templates/src/components/PDPTemplate/tokens/vitalPDPTemplate.ts
VancheeZze: Resolved

packages/vital-buttons/src/components/Buttons/tokens/vitalWTB.ts
```diff
@@ -39,8 +39,7 @@ const WhereToBuy = asButtonToken({
       vitalColor.TextPrimaryFooterCopy,
       vitalTextDecoration.Bold,
       vitalTextDecoration.Uppercase,
-      // @TODO: Create token? It should be same size for both mobile and desktop...
-      'text-m-base',
+      vitalTextDecoration.ButtonTextBase,
```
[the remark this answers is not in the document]
VancheeZze: Actually, it was a bad example of a token — deff not the Text Decoration token. So I when I decided to rename it and move to `vitalFontSize` token group, I saw there is already a `Base` token that has this class, so I just used that instead of creating a new one.

packages/__vital__/src/shadow/@bodiless/vital-templates/Page.ts
VancheeZze: Updated!

packages/vital-elements/src/components/Typography/tokens/vitalTypography.ts
VancheeZze: Heading tokens are now updated as discussed.

packages/vital-elements/src/components/Typography/tokens/vitalTypography.ts
```diff
@@ -122,52 +147,27 @@ const HeaderLink = as(
 
 const H1Link = asElementToken({
   ...H1,
-  Theme: {
-    _: as(
-      vitalTextDecoration.Bold,
-      HeaderLink,
-    ),
-  },
+  Theme: { _: as(HeaderLink) },
```
hvanyo, line 150: Can you actually remove H1Link, H2Link, ext.. and export HeaderLInk -- then find the use cases and define them as  

as(omit(H1, 'Theme'), HeaderLink)

VancheeZze: Good Call! Done!

packages/vital-link/src/components/Link/index.ts
VancheeZze: Done as well!

hvanyo (approved):

packages/vital-buttons/src/components/Buttons/tokens/vitalWTB.ts
```diff
@@ -39,7 +39,7 @@ const WhereToBuy = asButtonToken({
       vitalColor.TextPrimaryFooterCopy,
       vitalTextDecoration.Bold,
       vitalTextDecoration.Uppercase,
-      vitalFontSize.Base,
+      vitalFontSize.MobileBase,
```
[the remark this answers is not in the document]
VancheeZze: It does, yes. The `Base` token has `text-m-base lg:text-base` and the WTB button in baseline uses only `text-m-base`. Are we changing it so stat WTB uses `Base` token with two different font-size for mobile and desktop?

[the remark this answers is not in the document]
hvanyo: Well I think this is more of inconsistency is the design mockups is not following the typography styleguide (https://zeroheight.com/98a659433/p/1195ae-typographyfonts/b/5108bf) .   Design mockups do say that WTB is tablet/mobile = 14 px (baseline)  and Typography says buttons should be 14 mobile / 16 desktop (your PR)    So I would rather follow a pattern in the white label componets than add exceptions.   WTB button is out of scope for prototype anyway so this isn't super important.  I would say revert PR and will accept visual change diff on Desktop Buttons Size and not introduce a new token.

hvanyo (approved):

marcopagliarulo (approved):

PR #1988, opened 2023-02-22
[DO-NOT-MERGE] feat(backend): Convert Bodiless backend package to TypeScript

dewen:


packages/bodiless-backend/src/page.ts
```diff
@@ -12,25 +12,23 @@
...
-const backendFilePath = process.env.BODILESS_BACKEND_DATA_FILE_PATH || '';
-const backendStaticPath = process.env.BODILESS_BACKEND_STATIC_PATH || '';
-const IMG_ASSETS_PATH = `/images${path.sep}pages`;
-
-const getDirectories = (dir) => fs
+const getDirectories = (dir: string) => fs
   .readdirSync(dir)
   .filter((file) => fs.statSync(`${dir}/${file}`).isDirectory());
-// once we on node > 10.12.0
+// @todo: update to fs.mkdir - once we on node > 10.12.0
```
marcopagliarulo, line 29: We are actually using LTS version 16.9+, I think we can solve this todo in scope of this refactoring.

packages/bodiless-backend/src/backend.ts
```diff
@@ -0,0 +1,932 @@
...
+        'Access-Control-Allow-Headers',
+        'Origin, X-Requested-With, Content-Type, Accept',
+      );
+      res.header('Content-Type', 'application/json');
+      next();
+    });
+    this.setRoute(`${this.prefix}/changes`, this.getChanges);
+    this.setRoute(`${this.prefix}/changes/conflicts`, this.getConflicts);
+    this.setRoute(`${this.prefix}/get/commits`, Backend.getLatestCommits);
+    // this.setRoute(`${this.prefix}/change/amend`, this.setChangeAmend);
```
marcopagliarulo, line 309: I don't see any usage of the endpoints that you commented out, anyway I'd not remove these endpoints if we have any doubt. I'd suggest to discuss if we actually need these endpoints and take a clear decision.

packages/bodiless-backend/src/page.ts
hvanyo: @dewen  I would say for now, if its working lets descope update to fs.mkdir

packages/bodiless-backend/src/backend.ts
marcopagliarulo: @wodenx what do you think about these unused endpoints?

PR #1989, opened 2023-02-23
feat(gatsby-theme-bodiless): Allow disabling of editor in dev mode.

wodenx:
## Changes
Add a BODILESS_DISABLE_EDITOR env var which disables the editor even when running in development mode.

Will allow running a bodiless site using an alternative editor (eg Stackbit).

### DIFF
--- a/packages/gatsby-theme-bodiless/src/dist/Page.bl-edit.tsx
+++ b/packages/gatsby-theme-bodiless/src/dist/Page.bl-edit.tsx
@@ -104,6 +104,6 @@ const EditPage: FC<PageProps> = observer(({ children, ui, ...rest }) => {
   );
 });
 
-const Page = process.env.BODILESS_DISABLE_EDITOR ? StaticPage : EditPage;
+const Page = process.env.BODILESS_DISABLE_EDITOR === '1' ? StaticPage : EditPage;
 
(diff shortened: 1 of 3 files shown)


dewen (approved):

hvanyo (approved):

packages/gatsby-theme-bodiless/src/dist/Page.bl-edit.tsx
```diff
@@ -103,4 +104,6 @@ const Page: FC<PageProps> = observer(({ children, ui, ...rest }) => {
   );
 });
 
+const Page = process.env.BODILESS_DISABLE_EDITOR ? StaticPage : EditPage;
```
dewen, line 107: This might always be `true` if user set `process.env.BODILESS_DISABLE_EDITOR=0`

wodenx, line 107:
```suggestion
const Page = process.env.BODILESS_DISABLE_EDITOR === '1' ? StaticPage : EditPage;
```

wodenx: Fixed.

PR #1995, opened 2023-03-06
test: Add data-testid attributes for Cards page on vital demo site

uarlouski:
## Changes
todo


hvanyo:
@uarlouski  let's decline if you find that Chris's automated method of display component variations is suitable.

PR #2023, opened 2023-04-04
fix(@bodiless/search, gatsby-theme-bodiless): Fix/transpile to esm search and gatsby theme bodiless

### LINKED ISSUE #2025: Transpile @bodiless/search to ESM
## User Story
- AS A Developer
- I WANT @bodiless/search to be transpiled to ESM
- SO THAT @bodiless/search doesn't prevent a proper tree shaking due to its commonJS transpilation 

- AS A Developer
- I WANT A new global constant BL_IS_EDIT defined through definePlugin  
- SO THAT the packages can split use a simple condition for returning edit or static code. And Terser is able to remove dead code. 

- AS A Developer
- I WANT @bodiless/gatsby-theme-bodiless commonjs transpile configuration to be moved at package root
- SO THAT @bodiless/gatsby-theme-bodiless follow the common pattern for double transpilation in BodilessJs 

### Background/Description
The package must be transpiled to ESM, only the CLI scripts must be transpiled to CommonJS.

### Acceptance Criteria
src/scripts and src/searchTool are transpiled to commonJS and available inside the directories lib/cjs/scripts and lib/cjs/searchTool
A global boolean constant BL_IS_EDIT is available.

marcopagliarulo:
## Changes

- Transpile @bodiless/search to ESM, transpile to commonJS only the CLI scripts.
- Refactor @bodiless/gatsby-theme-bodiless transpile configuration to meet the standard agreed on BodilessJS for the packages requiring double transpile.
- Add a global constant BL_IS_EDIT through definePlugin which allow to remove dead code in base of the environment.

## Related Issues
Fixes #2025


packages/bodiless-bv/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "baseUrl": "."
   },
   "include": [
-    "src"
+    "src",
+    "../../@types/*"
```
wodenx, line 10: Can we put this into tsconfig.settings.json instead of repeating it for every package?

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
```diff
@@ -32,14 +32,14 @@ export const isStaticClientSide = !!(
   typeof window !== 'undefined'
   && window.document
   && window.document.createElement
-  && process.env.NODE_ENV === 'production'
+  && !BL_IS_EDIT
```
wodenx, line 35: Do we want to use this here or reserve it *only* for removign edit code?

wodenx (approved):

packages/bodiless-bv/tsconfig.json
marcopagliarulo: nope, because as by typescript documentation 
> It’s worth noting that files, include and exclude from the inheriting config file overwrite those from the base config file

Even if we define include into tsconfig.settings.json to be the most common case, then when transpilation is run, the path provided are updated relative to tsconfig.settings.json location, resulting in error because tsc then looks for files in ../../src

packages/bodiless-hydration/src/withoutHydration/withoutHydration.tsx
marcopagliarulo: You right, restored previous version

wodenx (approved):

PR #2031, opened 2023-04-17
feat(components, core): Add ability to move list items

wodenx:
## Changes
- Adds "move left" and "move right" buttons to local context menu for lists.
- Provides a mechanism to update local context menu position when unerlying component moves.
- Renders context menu only when visible (removes hidden context menus from dom).


packages/bodiless-core/src/PageEditContext/Store.tsx
```diff
@@ -43,6 +43,8 @@ export const defaultOverlaySettings: TOverlaySettings = {
  * Holds the current UI state for the editor.
  */
 export class PageEditStore implements PageEditStoreInterface {
+  @observable localContextMenuKey = new Date();
```
marcopagliarulo, line 46: Maybe it's unlikely, but if the localContextMenuKey is refreshed for two components within the same second, the generated key would be the same. Maybe getTime or generated a UUID could be safer

wodenx: We use localContextMenuKey.getTime() as the key, andt hat's ms since the epoch; https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime

marcopagliarulo (dismissed):

marcopagliarulo (dismissed):

PR #2050, opened 2023-04-26
feat: Upgrade/react18

marcopagliarulo:


packages/bodiless-layouts/src/SortableResizableWrapper/index.tsx
```diff
@@ -95,6 +119,7 @@ const SortableResizableWrapper = SortableElement((props: Props) => {
       enable={ENABLED_DRAG_SIDES}
       scale={1}
       className={className}
+      // @ts-ignore
```
marcopagliarulo, line 122: Fix the typescript issue here.

packages/bodiless-core/src/hoc.bl-edit.tsx
```diff
@@ -57,6 +57,7 @@ export const withContextActivator = (
  */
 export const withLocalContextMenu: HOC = Component => {
   const WithLocalContextMenu = (props: any) => (
+    // @ts-ignore
```
marcopagliarulo, line 60: Fix the typescript issue here.

PR #2051, opened 2023-04-26
feat(components, core): Allow moving list items.

wodenx:
## Changes

- Allows moving list items via buttons on local context menu
- Refactor local context menu to use floating-ui.  There is now a single local context menu which is positioned near the innermost element when a context is activated.


packages/bodiless-core/src/components/useLocalOptions.tsx
```diff
@@ -31,14 +26,14 @@ import { PageEditContextInterface } from '../PageEditContext/types';
...
  */
-const onPopupAlign = (domNode: Element) => {
-  const element = domNode as HTMLElement;
-  if (element.getBoundingClientRect().left < 0) {
-    element.style.visibility = 'hidden';
-  } else {
-    element.style.visibility = 'visible';
-  }
-};
+// const onPopupAlign = (domNode: Element) => {
```
dewen, line 29: Cleanup or add `@todo:`.

packages/bodiless-core/src/PageEditContext/Store.tsx
```diff
@@ -43,6 +43,8 @@ export const defaultOverlaySettings: TOverlaySettings = {
  * Holds the current UI state for the editor.
  */
 export class PageEditStore implements PageEditStoreInterface {
+  @observable localContextMenuKey = new Date();
```
dewen, line 46: Why not set this key as string?
```
new Date().getTime().toString();
```

packages/bodiless-core/src/components/ContextMenuItem.tsx
```diff
@@ -42,21 +42,29 @@ const ContextMenuItem = observer((props: IProps) => {
   const icon = option.icon ? (typeof option.icon === 'function' ? option.icon() : option.icon) : '';
   const title = typeof option.formTitle === 'function' ? option.formTitle() : option.formTitle;
   const description = typeof option.formDescription === 'function' ? option.formDescription() : option.formDescription;
-  const activateContext = option.activateContext
-    ? (typeof option.activateContext === 'function'
-      ? option.activateContext()
-      : option.activateContext)
-    : option.activateContext !== false;
+  // const activateContext = option.activateContext
```
dewen, line 45: cleanup?

dewen (approved):

marcopagliarulo (approved):

PR #2079, opened 2023-05-13
fix(bodiless-next): Fixing EmptySitemap error in @sites/--minimal-nex…

dewen:
…t-- build

### DIFF
--- a/packages/bodiless-next/src/Webpack/Sitemapxml.ts
+++ b/packages/bodiless-next/src/Webpack/Sitemapxml.ts
@@ -45,6 +45,7 @@ const generateSitemapXml = async (options: Partial<SitemapxmlConfig> | undefined
   if (!siteUrl) return;
 
   const pages = await getPages();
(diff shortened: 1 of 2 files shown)


packages/bodiless-next/src/Webpack/Sitemapxml.ts
```diff
@@ -68,7 +69,7 @@ const generateSitemapXml = async (options: Partial<SitemapxmlConfig> | undefined
     )
   )) : pages;
 
-  const serializedPages = filteredPages.filter(Boolean).map((page: string) => serialize(
+  const serializedPages = filteredPages.map((page: string) => serialize(
```
dewen, line 72: `filter` might not be necessary since the `pages` type is `string[]`

dewen: and empty string is a valid page item.

marcopagliarulo (approved):

PR #2087, opened 2023-05-18
chore: Complete Onboarding Intro Section

wodenx:
## Changes
- Add working code and some rewrites to "reaching-inside" lesson in onboard guide


packages/vital-elements/doc/GettingStarted/AnatomyOfAToken.md
```diff
@@ -10,8 +10,8 @@ additional styling and content.  It is equivalent to the following:
 ```ts
 const Welcome = asDialogToken({
   Theme: {
-    TitleWrapper: 'dialog-title-utility-classes',
-    MessageWrapper: 'dialog-message-utility-classes',
+    TitleWrapper: 'Dialog-title',
```
hvanyo, line 13: What are you trying to define here?   @wodenx   I know you had this originally and both confused me & Dewen, so I switched it..   To me this is defining the actual dialog title as a theme..  which doesn't make sense.   This naming convention  is not a pattern for taiwlind utlity class and normally best practice is create regular css classes with all lowercase -- although they are case insensitive.)

wodenx: @hvanyo  This part of the intro occurs before we introduce using tailwind and functional classes, so this is a semantic class name (the same one used in the original React docs example which we are transforming.  I can add a note to this effect in the first section.

- [ ] Add note about not using tailwind classes in early examples.

wodenx:
Reissuing to main

PR #2090, opened 2023-05-22
fix(filtering, page, next): editing bugs on NextJS

marcopagliarulo:

### DIFF
--- a/packages/bodiless-backend/src/page.ts
+++ b/packages/bodiless-backend/src/page.ts
@@ -441,7 +441,7 @@ class Page {
       const { sep, normalize } = path;
       const dirPath = normalize(this.directory);
       const [, pageRelativeDir] = dirPath.split(`${sep}data${sep}pages${sep}`);
-      if (!pageRelativeDir) {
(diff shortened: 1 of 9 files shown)


packages/bodiless-backend/src/page.ts
```diff
@@ -23,9 +23,17 @@ import Logger from './logger';
...
 
-const getDirectories = (dir: string) => fs
-  .readdirSync(dir)
-  .filter((file) => fs.statSync(`${dir}/${file}`).isDirectory());
+const getDirectories = (dir: string) => {
+  try {
+    return fs
+      .readdirSync(dir)
+      .filter((file) => fs.statSync(`${dir}/${file}`).isDirectory());
+  } catch (error) {
```
wodenx, line 31: What error are we ignoring here?

wodenx (approved):

packages/bodiless-backend/src/page.ts
marcopagliarulo: Here we check for subdirs, if there is any error, returns an empty array.

dewen (approved):

PR #2114, opened 2023-06-05
fix(bodiless-next, bodiless-navigation): Performance improvements

marcopagliarulo:
## Changes

1. Load google fonts on vital sites, next version, using next/font
2. Fix react-dom/server usage to prevent inclusion of legacy in production bundle
3. Stick to react-tag-autocomplete version to 7.0.0-rc.5 preventing breaking changes

### DM Sans font 
- Serve __vital_next__ and vital_demo_next sites
- Verify DM sans font is loaded and properly used
### DM Sans font 
- Serve any next site 
- Verify that framework.js doesn't include react-dom-server-legacy.browser.production.min.js


packages/bodiless-core/package.json
```diff
@@ -36,7 +36,7 @@
     "informed": "^4.44.0",
     "lodash": "^4.17.19",
     "rc-tooltip": "^5.1.1",
-    "react-tag-autocomplete": "^7.0.0-rc.5",
+    "react-tag-autocomplete": "7.0.0-rc.5",
```
wodenx, line 39: Should we just fix the breaking change?  I think it's just a type error.
We had an issue with the old package-lock getting deleted during a release (still unsure how it happened) which caused the upgrade.  Dewen restoerd the old pacakge lock, so we don't really need to pin, and can address the issue the next time we do an upgrade.

sites/__vital_next__/src/pages/_app.tsx
```diff
@@ -1,6 +1,39 @@
...
+import React from 'react';
+import type { AppProps } from 'next/app';
+import NextApp from '@bodiless/next/lib/NextApp';
+import { DM_Sans } from 'next/font/google';
+
 import '../css/style.css';
-// Add DM Sans font from Vital Demo
-import '--vital-next--/lib/dm-sans-font.css';
+
+const dm_sans = DM_Sans({
```
wodenx, line 8: Is there any way to encapsulate all this into @bodiless/next to make it easier to propagate to sites which already exist -- same for sijilar changes in the future.  The less we have to change int he starter the better, esp. as we build more sites.

wodenx (commented):
See inline comments

sites/__vital_next__/src/pages/_app.tsx
marcopagliarulo: I tried it, but next/font doesn't allow it

PR #2144, opened 2023-06-30
JEPZ-496 Vital Dividers

salikn:


packages/vital-divider/src/components/Divider/DividerClean.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { FC, Fragment } from 'react';
```
hvanyo, line 1: @bodiless/vital-divider: src/components/Divider/DividerClean.tsx(1,21): error TS6133: 'Fragment' is declared but its value is never read.
@bodiless/vital-divider: ERROR: "build:lib" exited with 2.

PR #2148, opened 2023-07-01
feat(vital-card): Add Product Card

VancheeZze:
## Changes

- Added new `ProductCard` component with tokens corresponding to Figma Design variations.
- Added examples of `ProductCard` to Knapsack `Vital 2.0` Page.

### Dev Notes:
- The PR is based off of the previous Token work that is not yet merged.


wodenx:
I think i like separating the product card into it's own `/src/components/ProductCard` directory - it will allow shadowing the product card independently from other cards - and this could even be moved to the vital-products package when that is available.  However, all new `src/components` should be created using the component scaffolder to ensure consistency and save time.
Also we had discussed adding a `/src/componetns/ProductCardElements` token collection holding all the "component-level" design tokens used by the product card, rght?

VancheeZze:
> I think i like separating the product card into it's own `/src/components/ProductCard` directory - it will allow shadowing the product card independently from other cards - and this could even be moved to the vital-products package when that is available. However, all new `src/components` should be created using the component scaffolder to ensure consistency and save time. Also we had discussed adding a `/src/componetns/ProductCardElements` token collection holding all the "component-level" design tokens used by the product card, rght?

Sure — I started it as a new component, but then saw the stub of the product card and moved it all there. Will move it to it own folder with the scaffolder script! Also for the component-level tokens — these were created in a previous PR here: https://github.com/johnsonandjohnson/Bodiless-JS/pull/2126/files#diff-8ee8b1cb6e5a7c41e7a576474d2c00ac7c10c0ea2f16474d5dcb5fdbc60e4612

packages/vital-card/src/components/ProductCard/tokens/vitalProductCard.ts
```diff
@@ -0,0 +1,216 @@
...
+  Components: {
+    DescriptionWrapper: replaceWith(P),
+    Description: on(EditorPlainClean)(vitalEditorPlain.Default),
+  },
+  Content: {
+    Description: withPlaceholder('Description Text'),
+  }
+});
+
+/**
```
hvanyo, line 78: @VancheeZze    These buttons are not intended to be another slot and not reuse CTALINK (which will go to the /pdp)
AC:  This buttons behavior needs to be clickable and could have a different link than to the PDP.   (NOTE assume IT WILL HAVE) 
So I think in Cards we need another slot to support button.   ButtonWrapper/ ButtonLink / Button Text.   This could be pplied to generic card or we start with new productclean card.

hvanyo (dismissed):

samlynndavis (dismissed):

samlynndavis (approved):

KaterinaPashkevich (dismissed):

samlynndavis (dismissed):

KaterinaPashkevich (approved):

samlynndavis (approved):

PR #2172, opened 2023-07-17
Option1:   Skip defining vitalColors and use tailwind classes directly

hvanyo:


packages/vital-button/src/components/ButtonTokens/index.ts
```diff
@@ -22,31 +22,34 @@ const vitalButtonElement = asTokenGroup(ButtonMeta)({
   ShadowButtonFocus: 'focus:drop-shadow-button',
```
wodenx, line 22: this file is in the worng place, right?  It should be
`vital-elements/src/components/ButtonElement/tokensvitalButtonElement.ts`

PR #2194, opened 2023-07-27
test: Add Lighthouse performance checks

uarlouski:

### DIFF
--- a/.github/workflows/build-test-lint-simple.yml
+++ b/.github/workflows/build-test-lint-simple.yml
@@ -59,6 +59,27 @@ jobs:
           name: Accessibility
           path: accessibility-reports/
           retention-days: 30
+  
+  lhci:
+    name: Lighthouse
+    if: ${{ github.event_name == 'schedule' || github.event.label.name == 'ready-to-test' }}
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event_name == 'schedule' && 'main' || github.event.pull_request.head.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test-lint-simple.yml
```diff
@@ -59,6 +59,27 @@ jobs:
...
           retention-days: 30
+  
+  lhci:
+    name: Lighthouse
+    if: ${{ github.event_name == 'schedule' || github.event.label.name == 'ready-to-test' }}
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event_name == 'schedule' && 'main' || github.event.pull_request.head.sha }}
```
dewen, line 70: `&& 'main'` doesn't make sense for me here, typo?

lighthouserc.js
```diff
@@ -0,0 +1,39 @@
+// https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md
```
dewen, line 1: can we move `lighthouserc.js` to `scripts` folder?

.github/workflows/build-test-lint-simple.yml
uarlouski: that's an analog of ternary operator, if the event name is 'schedule' we checkout main branch, otherwise we checkout current PR SHA

dewen: i see it's an assignment, sorry mess with condition check

dewen (approved):

PR #2197, opened 2023-08-09
test: Disable tests for Accordion page

dtargons:


hvanyo:
@dtargons  thank you, I am not going to moveforward with breaking accordions and fixing peformance that way..

PR #1648, opened 2022-04-26
feat(tokens, vital-demo): Storybook 

wodenx:
## Changes
Adds storybook support for vital  DS.

PR #2202, opened 2023-08-15
Feature/token demo 2023

hvanyo:
## Changes
Updated: https://github.com/johnsonandjohnson/Bodiless-JS/pull/1648

## Test Instructions
Storybook (run from vital-demo)

- npm run storybook, test tokens on typogorapy, font-size and header pages.
- some token combinations don't make sense, this is expected
- On header page - content is editable and persisted to session storage, but image uploads are not possible


packages/bodiless-tokens/src/TokenMap.ts
```diff
@@ -12,70 +12,125 @@
...
+    const tokens = Object.keys(collection).reduce(
+      (toks, key) => ({
+        ...toks,
+        [key]: as(collection[key] as Token),
+      }),
+      {}
+    );
+    this.add(tokens);
+    return this;
+    // TBD Conflict resolution
```
hvanyo, line 57: @salikn  -- 
Lines 59 - 62 was what was on MAIN a year ago.
Lines 64 - 66 was is on MAIN today.   From what I can tell we update the token to HOCWithMEta to access token?.meta? 
Chris replaced addCollection  (lines 47-56)   
that all seems to work..  and will delete all these lines but I think its useful to see and sets up for my question.

packages/bodiless-tokens/src/TokenMap.ts
```diff
@@ -12,70 +12,125 @@
...
+} from '@bodiless/fclasses';
 
 export type Tokens = {
   [key: string]: HOC,
 };
 
-class TokenMap<P> {
-  protected map = new Map<string, HOC>();
+const extractMeta = (token?: Token): TokenMeta => {
+  // TODO  error TS2339: Property 'meta' does not exist on type 'HOC<{}, {}, {}>'.
```
hvanyo, line 24: @salikn  
Line 25 gives type error in line 24 comment.  
I played with switching token?: HOCWithMeta  but it really messes up a lot of other things so not sure what was right move.  
So not sure how to fix type error.   Ideas?

And obviosuly I would delete line 26 once I fix type error.

packages/bodiless-tokens/src/TokenMap.ts
```diff
@@ -12,70 +12,125 @@
...
+  // if (typeof token === 'function') return token?.meta || {};
+  if (typeof token === 'function') return {};
+  if (typeof token === 'string') return {};
+  return (token as TokenSpec<any, any>)?.Meta || {};
```
hvanyo, line 28: @salikn  correct way to fix this type error? 
(Typescript in javascript is not my expertise and I just find it PITA. )

PR #2204, opened 2023-08-16
Feat/pdpcarousel

hvanyo:


packages/vital-carousel/src/components/utils/sliderSimpleInit.ts
```diff
@@ -62,6 +71,14 @@ export const sliderSimpleInit = (sliderSimpleElement: HTMLElement) => {
       const NumThumbs = 4;
       for (const button of Array.from(buttonsThumbs)) {
```
VancheeZze, line 72: Shouldn't this code be under the `if (sliderThumbCarousel.length !== 0) {...}` condition above? There is no need to add controls if there are no thumbs, right? Or is it still has to present? 

Also if the condition above is not met, then this line might error since `buttonsThumbs` will be undefined in `Array.from(buttonsThumbs)`

`Array.from(buttonsThumbs)` could also be saved to a var and reused for a tiny speed bump: https://github.com/johnsonandjohnson/Bodiless-JS/pull/2204/files#diff-6c6a78fcbd165e433925aac71d373a5d6284eabbdf275a4825969bc59913afe7R52.

packages/vital-product/src/components/PDPTemplate/tokens/vitalPDPTemplate.ts
```diff
@@ -103,14 +117,15 @@ const Default = asPDPTemplateToken({
   },
   Layout: {
     ContentWrapper: 'flex flex-wrap',
-    ProductImageWrapper: 'flex justify-center w-full lg:w-1/2',
+    ProductCarouselWrapper: 'lg:w-1/2',
```
VancheeZze, line 120: Should this `ProductCarouselWrapper` be hidden on mobile and only be visible on desktop — opposite from the `MobileProductCarouselWrapper` below?

VancheeZze (approved):

packages/vital-carousel/src/components/utils/sliderSimpleInit.ts
hvanyo: @VancheeZze  I think this code is under the condition in line 46.
The typescript version only runs on edit so performance doesn't matter.
The non-typescript javascript version runs on static and reuses buttonsThumbs array with typescript cast.  
I learned the lesson the hardway that browser doesn't like to run typescript'ed javascript :)

packages/vital-product/src/components/PDPTemplate/tokens/vitalPDPTemplate.ts
hvanyo: I was letting component tokens in line 92 / 99 do the responsive hiding.  I had to add the lg:hidden on line 121 though otherwise it screwed up the flex box  lg w-1/2 on desktop.   As i made the change,  I pondered if the way we should do this was by wrapper on template or by the reponsive token..  Opinions?

PR #2209, opened 2023-08-21
feat(gtm): update GTM snippet

dewen:


sites/vital-demo-next/.env.site
```diff
@@ -77,3 +77,6 @@ BODILESS_GENERATED_DESTINATION_PATH='generated'
 BODILESS_STATS_PATH='./public/generated'
 BODILESS_BUILD_STATS=1
 # BODILESS_OPEN_STATS=1
+
+GTM_ID='GTM-WSPHT2V'
+GTM_DATA_LAYER='dataLayer'
```
marcopagliarulo, line 82: NextJS strip the environment variables from frontend unless prefixed by NEXT_ or added to the public environment variables in nextJS config, @bodiless/next preconfigure the public env variables here https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/bodiless-next/src/NextConfig/getPublicEnv.ts#L34

I suggest to add those two variables to the list

sites/vital-demo-next/src/pages/_document.tsx
```diff
@@ -1,35 +1,16 @@
...
-    <noscript
-      key="google-tagmanager-noscript"
-      dangerouslySetInnerHTML={{
-        __html: `<iframe src="https://www.googletagmanager.com/ns.html?id=${id}" height="0" width="0" style="display: none; visibility: hidden" aria-hidden="true"></iframe>`
-      }}
-    />
-  );
+  const id = process.env.GTM_ID || 'GTM-XXXXXX';
+  const dataLayerName = process.env.GTM_DATA_LAYER || 'dataLayer';
+  documentProps.PostHead = generateGTMScript(id, dataLayerName);
```
marcopagliarulo, line 12: instead of adding the script in this way, why not create an HOC in @bodiless/next which check if the env variable exists, generates the script and add it to the page using React Helmet, like we do for Canonical URL? https://github.com/johnsonandjohnson/Bodiless-JS/blob/main/packages/bodiless-next/src/Page.static.tsx#L50

In this way the setup at site level would be simplest, just define the env variables and optionally add the no script tag when required

sites/vital-demo-next/package.json
```diff
@@ -22,6 +22,7 @@
     "@bodiless/fclasses": "^1.0.0-rc.42",
     "@bodiless/ga4": "^1.0.0-rc.42",
```
hvanyo, line 23: @dewen  why readd the @bodiless/ga4 ? its at line 23 & line 25

marcopagliarulo (approved):

PR #2211, opened 2023-08-21
[WIP] feat(vital-card): JEPZ-408 | Hero Card 2.0

samlynndavis:


packages/vital-card/src/components/HeroCard/tokens/vitalHeroCard.ts
```diff
@@ -0,0 +1,294 @@
...
+      Body: withPlaceholder('BUTTON'),
+    },
+  }
+);
+
+/**
+ * Tertiary Hero Card Button.
+ * Plain vital button with `WithTertiaryStyle` token and `READ MORE` placeholder.
+ */
+const TertiaryHeroCardButton = asButtonToken(
```
VancheeZze, line 54: I'd probably combine these cards, they have a lot in common. Something like this:

```js
const PlainButton = asButtonToken(vitalButton.Plain, {
    Content: {
      Body: withPlaceholder('BUTTON'),
    },
});

// And then

const PrimaryHeroCardButton = asButtonToken(PlainButton, vitalButton.WithPrimaryStyle),
const SecondaryHeroCardButton = asButtonToken(PlainButton, vitalButton.WithSecondaryStyle),
// ...
```

packages/vital-card/src/components/HeroCard/tokens/vitalHeroCard.ts
```diff
@@ -0,0 +1,294 @@
...
+  ...WithHorizontalOrientationBase,
+  Layout: {
+    Wrapper: 'sm:gap-16px md:gap-24px',
+  },
+  Spacing: {
+    ContentWrapper: 'px-16px',
+    ImageWrapper: 'px-16px md:px-0px',
+  },
+});
+const HorizontalWithImageLeft = asHeroCardToken(
```
VancheeZze, line 98: I think these orientational tokens should also be named with `WithHorizontalWithImageLeft ` and `With HorizontalWithImageRight`.

I'd also keep the names consistent. We have `WithHorizontalRightOrientation` exported from base. I thought keeping the names of these tokens may help us in future. What I did for some other cards is changing the names of imports like

```js
import Base, {
  WithHorizontalLeftOrientation as WithHorizontalLeftOrientationBase,
  WithHorizontalRightOrientation as WithHorizontalRightOrientationBase,
  // ...
} from '../../Card/tokens/Base';
```

and then in this file define `WithHorizontalLeftOrientation` and `WithHorizontalRightOrientation` that will be part of the token collection. 

This way we would have consistent token names across multiple cards

packages/vital-card/src/components/HeroCard/tokens/vitalHeroCard.ts
```diff
@@ -0,0 +1,294 @@
...
+  },
+  Theme: {
+    SubtitleWrapper: vitalTypography.HeadlineMedium,
+  },
+  Spacing: {
+    SubtitleWrapper: vitalSpacing.MarginBottomSmall,
+  }
+});
+
+const WithNoSubtitle = asHeroCardToken({
```
VancheeZze, line 191: I think we were trying to avoid `WithNo...` tokens. cc @hvanyo . Why would we have both `WithSubtitle` and `WithNoSubtitle` tokens.

VancheeZze (changes requested):

packages/vital-card/src/components/HeroCard/tokens/vitalHeroCard.ts
samlynndavis: I forgot that this slot isn't part of the component by default when I was quickly writing all of the `WithNo...`  tokens. 

I'll remove this one.

However, the AC _does_ ask for a token to remove both the description and image (both default slots).

